PR #2, opened 2026-04-09
feat: add core MCP server, TrainerClient tools and Resources

### LINKED ISSUE #7: Core MCP server framework and TrainerClient tools
### Description

Implement the core MCP server and all TrainerClient-backed tools per [KEP-936](https://github.com/kubeflow/community/tree/master/proposals/936-kubeflow-mcp-server).

**Core Framework (`kubeflow_mcp/core/`)**
- Server factory with persona-based tool filtering
- Security: K8s name validation, training bounds, secret masking, namespace enforcement, heuristic script safety
- Auth: Bearer token + JWT/OIDC for HTTP transport
- Policy: 4 personas (`readonly`, `data-scientist`, `ml-engineer`, `platform-admin`) with allow/deny overrides via `~/.kf-mcp-policy.yaml`
- Resilience: Thread-safe rate limiter and circuit breaker
- Structured JSON audit logging with correlation IDs
- Config system (YAML + env-var fallback chain)

### Use Case

This implementation is gonna be the testable baseline for this new repo - after this implementation reviewers can directly test this mcp project right away thoroughly to test workflows or even make inspector against a real working server.

```
kubeflow-mcp serve \
  --clients trainer \             # modules: trainer, optimizer (stub), hub (stub)
  --persona ml-engineer \         # readonly | data-scientist | ml-engineer | platform-admin
  --mode full \                   # full | progressive | semantic
  --transport stdio \             # stdio | http | sse
  --auth-token SECRET \           # bearer token for HTTP auth (dev/staging)
  --log-level INFO \              # DEBUG | INFO | WARNING | ERROR
  --log-format console \          # console | json (auto-detected if omitted)
  --no-banner
```

### Proposed Solution

**CLI & MCP Resources**
- `serve` command with `stdio` / `http` / `sse` transport
- `--persona`, `--auth-token`, `--instruction-tier`, `--mode` flags
- `status` and `agent` subcommands
- Resources: Training patterns, platform guides, troubleshooting

**Tests**
- CLI tests — flags, config fallback, auth/resilience wiring
- SDK contract tests — dataclass fields, enum values, API signatures
- Architecture tests — tool metadata, persona gating, resource registration

### Component

Core

abhijeet-dhumal:
## Description

Implements the core MCP server framework and all TrainerClient-backed tools per [KEP-936](https://github.com/kubeflow/community/tree/master/proposals/936-kubeflow-mcp-server).
Claude plugin and skills moved to a separate PR per reviewer feedback.

### Core Framework (`kubeflow_mcp/core/`)

- **Server factory** with persona-based tool filtering, and progressive/semantic meta-tool modes
- **Config** — YAML config file with env-var fallback chain
- **Security** — K8s name validation, training bounds, secret masking, namespace enforcement, heuristic script safety
- **Auth** — Bearer token + JWT/OIDC for HTTP transport
- **Policy** — Persona access control (`readonly`, `data-scientist`, `ml-engineer`, `platform-admin`) with allow/deny overrides
- **Resilience** — Thread-safe rate limiter and circuit breaker
- **Logging** — Structured JSON audit logging with correlation IDs

### Test Suite (131 tests)

- **`cli_test.py`** (25) — CLI flags, config fallback, auth/resilience wiring, SSE transport
- **`sdk_contracts_test.py`** (68) — SDK dataclass fields, enum values, API signatures, K8s client methods
- **`test_architecture.py`** (38) — Tool metadata, persona gating, instruction tiers, resource registration

## Checklist

- [x] Tests pass locally (`make test-python`) — 131 passing
- [x] Linting passes (`make verify`) — 0 errors
- [x] Documentation updated

## Related Issues

Implements [KEP-936](https://github.com/kubeflow/community/tree/master/proposals/936-kubeflow-mcp-server)


abhijeet-dhumal:
Hey @andreyvelich @astefanutti @kramaranya @szaher @Electronic-Waste - the 2nd onboarding phase is ready for review now 🏁 
Would appreciate a review when you get a chance 🙏

kramaranya:
Thanks @abhijeet-dhumal, is it possible to split the changes into a few PRs (for example MCP server, tools, docs/plugin/benchmark)?

abhijeet-dhumal:
> Thanks @abhijeet-dhumal, is it possible to split the changes into a few PRs (for example MCP server, tools, docs/plugin/benchmark)?

Thanks @kramaranya .. I completely understand the preference for smaller PRs, and I normally follow that approach.

The context here is that after discussing with @andreyvelich and @thesuperzapper, we agreed on a phased onboarding strategy with a time constraint: land the project as a cohesive unit first, then refine via focused follow-up PRs. I tried to keep commits in this PR structured as logically separate components to make review easier.

Splitting into parallel PRs now is doable, but these components have tight interdependencies (e.g. tools depend on core server, benchmarks test the tools, security hardening touches multiple layers), so it would add significant overhead to manage the PR dependency chain and likely slow down the merge timeline.

@andreyvelich @astefanutti  @thesuperzapper  What way do you recommend, Happy to adjust either way.

andreyvelich:
@abhijeet-dhumal Maybe you can separate Claude plugin and SKILLS into separate PR? So we can review MCP tools first.

abhijeet-dhumal:
> @abhijeet-dhumal Maybe you can separate Claude plugin and SKILLS into separate PR? So we can review MCP tools first.

Thanks @andreyvelich - done ✅  
I've moved plugin.json and the 5 Claude Code skills into a separate branch and will open a follow-up PR for those once the MCP tools land. ROADMAP.md is updated to reflect the split. 
This PR now focuses purely on the core MCP server, TrainerClient tools, and tests.

kubeflow_mcp/trainer/api/lifecycle.py
```diff
@@ -0,0 +1,193 @@
...
+    Warning:
+        Requires confirmation in MCP clients - accidental deletion prevented.
+    """
+    try:
+        ns_err = check_namespace_allowed(namespace)
+        if ns_err:
+            return ns_err.model_dump()
+
+        client = mcp_utils.get_trainer_client()
+        client.delete_job(name=name)
```
jaiakash, line 95: `delete_training_job` ignores the namespace parameter — uses default namespace regardless of what's passed. Can we check this!
Fix: Use training.py's `_get_client(namespace)` pattern, or pass namespace to `delete_job()`.

kubeflow_mcp/common/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    Call ``reset_clients()`` to force re-creation (e.g. after kubeconfig change).
+    """
+    from kubernetes import client, config
+
+    config.load_config()
+    configuration = client.Configuration.get_default_copy()
+    configuration.retries = 1
+    configuration.socket_options = None  # rely on OS defaults
+    api_client = client.ApiClient(configuration)
+    api_client.rest_client.pool_manager.connection_pool_kw["timeout"] = K8S_TIMEOUT
```
jaiakash: This reaches deep into urllib3 internals via the kubernetes client. It works today but can break silently on kubernetes package upgrades. 
Can we consider using Configuration.connect_timeout and Configuration.read_timeout instead ?

kubeflow_mcp/core/resilience.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class CircuitState(Enum):
+    CLOSED = "closed"
+    OPEN = "open"
+    HALF_OPEN = "half_open"
+
+
+@dataclass
+class CircuitBreaker:
```
jaiakash, line 41: RateLimiter correctly uses `threading.Lock`, but CircuitBreaker does not. In an HTTP transport scenario (threaded server), concurrent tool calls will race on state transitions, potentially leaving the breaker in an inconsistent state (e.g., never closing after recovery).

Fix: Add a `threading.Lock` to `CircuitBreaker`

kubeflow_mcp/trainer/api/training.py
```diff
@@ -0,0 +1,850 @@
...
+def _get_client(namespace: str | None = None) -> Any:
+    """Return a TrainerClient targeting the given namespace.
+
+    When *namespace* is ``None`` the shared singleton (default kubeconfig
+    namespace) is returned.  When a namespace is explicitly requested a
+    fresh ``TrainerClient`` is created so the SDK sends the TrainJob to
+    that namespace.
+    """
+    if namespace is None:
+        return get_trainer_client()
```
jaiakash, line 136: When namespace is not None, a new TrainerClient is created on every call.

Fix: Can we reuse the shared client and pass namespace at call site.

jaiakash:
Hi @abhijeet-dhumal Thanks for a ton for this.

I’ve left comments on a couple of issues above—could you please take a look at those? Aside from that, everything looks good to me at this stage of the project. 😊

Disclaimer: Used Claude Opus for PR review.

<details>
<summary>Prompt Details</summary>

Please perform a high-level code review of this PR against the main branch.
Since this is an early-stage project with a large volume of changes, do not focus on line-by-line nitpicks. Instead, prioritize:

Ensuring the PR description is accurately implemented
Verifying that all intended features are correctly and completely delivered
Identifying any security risks or vulnerabilities
Checking for violations of coding standards, best practices, or architectural guidelines
Highlighting any potential scalability, maintainability, or design concerns
Calling out missing tests, validation, or error handling where relevant

Focus on important issues and risks, not minor style suggestions.
Provide actionable feedback and clearly categorize findings by severity (e.g., critical, major, minor).

Also point out anything that could cause production issues, even if it is not explicitly part of the PR scope.

</details>

kubeflow_mcp/core/resilience.py
abhijeet-dhumal: Yep, spotted this during internal review. 
I have updated it accordingly, CircuitBreaker now has a threading.Lock on all state transitions. Also tightened the half-open path - slot reservation is atomic inside can_execute() so concurrent threads can't exceed the probe count.

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: Right, that was wasteful. Moved to a shared thread-safe cache (get_trainer_client_for_namespace) in common/utils.py with FIFO eviction at 64 entries. Both training.py and lifecycle.py use it now.

kubeflow_mcp/common/utils.py
abhijeet-dhumal: Agreed, that was fragile. Now switched to Configuration.connect_timeout / Configuration.read_timeout - much safer across kubernetes client upgrades.

kubeflow_mcp/trainer/api/lifecycle.py
abhijeet-dhumal: Good catch - this was on my list too. I have addressed it now: delete_training_job now routes through get_trainer_client_for_namespace(namespace) so the explicit namespace is respected. Response also includes the resolved namespace for visibility.

abhijeet-dhumal:
> Hi @abhijeet-dhumal Thanks for a ton for this.
> 
> I’ve left comments on a couple of issues above—could you please take a look at those? Aside from that, everything looks good to me at this stage of the project. 😊
> 
> Disclaimer: Used Claude Opus for PR review.
> 
> Prompt Details

Thanks @jaiakash - all 4 items addressed in the latest push.. Appreciate the thorough review 🙌  
The CircuitBreaker race and urllib3 internals catches were solid finds

abhijeet-dhumal:
> @abhijeet-dhumal Maybe you can separate Claude plugin and SKILLS into separate PR? So we can review MCP tools first.

@andreyvelich @kramaranya @astefanutti Apologies for the size of this PR - I understand it makes review harder.
IMHO splitting this PR further would require stubs or dead code in each PR just to pass CI, without making review easier.
This PR is the testable baseline for this new repo - reviewers can directly test this mcp project right away thoroughly to test workflows or even make inspector against a real working server. 
```
kubeflow-mcp serve \
  --clients trainer \             # modules: trainer, optimizer (stub), hub (stub)
  --persona ml-engineer \         # readonly | data-scientist | ml-engineer | platform-admin
  --mode full \                   # full | progressive | semantic
  --transport stdio \             # stdio | http | sse
  --auth-token SECRET \           # bearer token for HTTP auth (dev/staging)
  --log-level INFO \              # DEBUG | INFO | WARNING | ERROR
  --log-format console \          # console | json (auto-detected if omitted)
  --no-banner
```

Re: why MCP Resources are included
The 3 resource guides (training-patterns, platform-fixes, troubleshooting) are not documentation — they're part of the tool runtime. Resources ARE packaged in the MCP server - they get bundled with pip install, and are registered on the server at startup via FastMCP's mcp.resource(uri). Any MCP client that connects gets access to them through the standard resources/read protocol method.

MCP Resources are a protocol-level feature that agents read on-demand when tools reference them:
- pre_flight() and estimate_resources() return trainer://guides/platform-fixes when OpenShift is detected
- fine_tune() returns trainer://guides/training-patterns when the user needs a LoRA script alternative
- get_training_logs() returns trainer://guides/troubleshooting when OOM/CUDA errors are detected
- Server instructions reference all 3 URIs to guide agent behavior

Without these, tool responses would contain issues that agents can't resolve. They're co-located with the tools because they're registered by the same client module and tested together (architecture tests verify all URIs are resolvable).

Happy to clarify anything further — let me know if you'd like me to adjust the scope.

kubeflow_mcp/trainer/api/lifecycle.py
```diff
@@ -0,0 +1,211 @@
...
+    if name_err:
+        return name_err.model_dump()
+
+    try:
+        ns_err = check_namespace_allowed(namespace)
+        if ns_err:
+            return ns_err.model_dump()
+
+        ns = mcp_utils.get_trainer_effective_namespace(namespace)
+        persona = os.getenv("KUBEFLOW_MCP_PERSONA", "readonly")
```
kramaranya: This reads persona from `KUBEFLOW_MCP_PERSONA`, but the server resolves persona from CLI/config and passes it into `create_server()` without guaranteeing this env var is set.

That means a server started as `--persona platform-admin` via CLI flag or config can still behave like `"readonly"` here and reject operations on non-MCP-managed jobs, even though the tool was already exposed by persona filtering.

Could we pass the effective persona through server/module state instead of re-reading the env var here? Same issue appears in `update_training_job()` as well.

---
_AI-assisted review — verified before posting_

kubeflow_mcp/core/server.py
```diff
@@ -0,0 +1,379 @@
...
+    allowed_funcs = [f for f in all_tool_funcs if f.__name__ in final_allowed]
+
+    logger.debug(f"Final allowed tools after policy: {len(final_allowed)}")
+
+    # --- Mode dispatch ---
+    valid_modes = ("full", "progressive", "semantic")
+    if mode not in valid_modes:
+        raise ValueError(f"Invalid mode '{mode}'. Must be one of: {', '.join(valid_modes)}")
+
+    if mode in ("progressive", "semantic"):
```
kramaranya, line 346: In progressive/semantic mode, only the meta-tools are wrapped by `_audit_wrap()`. `execute_tool()` then calls the raw underlying tool from `TOOL_REGISTRY`, so infrastructure failures are recorded against the breaker for `execute_tool`, not the actual tool that failed.

In practice, a few failures in one proxied tool can open the breaker for all proxied tools behind `execute_tool`, and the audit logs also become much less useful because everything is attributed to the meta-tool.

Would it make sense to store audited/per-tool-wrapped callables in `TOOL_REGISTRY`, or otherwise plumb the underlying tool name through the audit/breaker path?

---
_AI-assisted review — verified before posting_

kubeflow_mcp/trainer/api/training.py
```diff
@@ -0,0 +1,1196 @@
...
+    if func_args:
+        params = ", ".join(f"{k}=None" for k in func_args)
+        wrapped = f"def {func_name}({params}):\n"
+    else:
+        wrapped = f"def {func_name}():\n"
+    for line in lines:
+        wrapped += f"    {line}\n"
+
+    script_dir = _get_script_dir()
+    script_path = os.path.join(script_dir, "_mcp_training_script.py")
```
kramaranya: `_make_train_func()` always writes to the same `_mcp_training_script.py` path under the shared temp directory.

With HTTP/SSE transports enabled, concurrent requests can race here: one request can overwrite the file while another is still relying on that path for `inspect`/source extraction, which can lead to the wrong script being associated with the wrong submission.

Could this use a unique filename per invocation instead, for example a UUID or `NamedTemporaryFile`?

---
_AI-assisted review — verified before posting_

kubeflow_mcp/trainer/api/training.py
```diff
@@ -0,0 +1,1196 @@
...
+    namespace is explicitly provided a cached scoped client is returned.
+    """
+    if namespace is None:
+        return get_trainer_client()
+    return get_trainer_client_for_namespace(namespace)
+
+
+def _inject_ownership_label(options: list) -> list:
+    """Append the MCP ownership label to the options list."""
+    options.append(Labels(labels={MCP_MANAGED_LABEL: MCP_MANAGED_VALUE}))
```
kramaranya, line 146: `_build_runtime_patch()` appends `Labels(labels=labels)` when callers provide labels, and later `_inject_ownership_label()` appends a second `Labels(...)` option for `kubeflow-mcp/managed-by`. In the current SDK, `Labels.__call__()` assigns `metadata["labels"] = self.labels`, so the later option overwrites the earlier one instead of merging.

That means MCP-managed jobs lose any caller-provided labels. It would be safer to merge the ownership label into the caller label dict up front and emit a single `Labels(...)` option.

---
_AI-assisted review — verified before posting_

kubeflow_mcp/trainer/api/lifecycle.py
abhijeet-dhumal: Good catch - I think this was a leftover from early prototyping when persona was only set via env var.
Let me check

kubeflow_mcp/core/server.py
abhijeet-dhumal: Fair point. The intent was to keep meta-tools lightweight, but you're right that attributing all failures to a single execute_tool breaker defeats the purpose.

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: This wasn't an issue in stdio mode (single client, sequential calls), but you're right it becomes a race in HTTP/SSE with concurrent requests. 
let me check

abhijeet-dhumal:
Thanks for the thorough review, @kramaranya ! I have resolved the comments raised..
Looking forward to the next round whenever you get a chance. 😄

kubeflow_mcp/core/server.py
priyank766: Looks like the round-1 fix only closes half of this. `execute_tool` now calls `get_breaker(tool_name)` for the underlying tool, but `TOOL_REGISTRY[name]["func"]` still holds the raw function - `_audit_wrap` never runs on the actual call.

In `progressive` / `semantic` mode that means:

1. **Rate limiting is skipped** - `_rate_limiter.acquire()` only gates the meta-tool, not the proxied call.
2. **Audit logs are misattributed** - entries record `tool="execute_tool"` with the meta-tool's args, so `mask_sensitive_data` runs over `{tool_name, arguments}` instead of the real parameters. Secret masking on the actual call is effectively bypassed.
3. **`_inject_meta` is skipped**, so response shape differs between modes.
4. **Outer breaker aggregates across tools** - `_audit_wrap` on `execute_tool` still feeds a single `"execute_tool"` breaker. Five unrelated tools failing once each trips the whole meta-tool path, re-introducing the cross-tool aggregation the inner per-tool breaker was meant to prevent.

Storing the audited callable in the registry would close all four at once:

```python
# server.py, before init_dynamic_tools(...)
audited_funcs = [_audit_wrap(f) for f in tool_funcs]
init_dynamic_tools(audited_funcs, all_descriptions)
```

The duplicated breaker block inside `execute_tool` can then drop, and `execute_tool` itself probably wants a sentinel breaker name (or no outer `_audit_wrap`) so it doesn't share the per-tool pool.

_AI-assisted review — verified before posting_

kubeflow_mcp/core/dynamic_tools.py
```diff
@@ -0,0 +1,404 @@
...
+
+    return {"tools": results}
+
+
+def _is_infrastructure_error(result: Any) -> bool:
+    """Return True if a tool result indicates a K8s/SDK infrastructure failure."""
+    if not isinstance(result, dict):
+        return False
+    code = result.get("error_code", "")
+    return code in (ErrorCode.KUBERNETES_ERROR, ErrorCode.SDK_ERROR, ErrorCode.TIMEOUT)
```
kramaranya: I think it's duplicated in server.py, let's move it to common

kubeflow_mcp/core/server.py
```diff
@@ -0,0 +1,383 @@
...
+    # Collect all available tools from loaded modules + health tools
+    all_tool_funcs: list = []
+    all_tool_names: set[str] = set()
+    for module in loaded_modules.values():
+        for tool_func in getattr(module, "TOOLS", []):
+            all_tool_names.add(tool_func.__name__)
+            all_tool_funcs.append(tool_func)
+    for tool_func in HEALTH_TOOLS:
+        all_tool_names.add(tool_func.__name__)
+        all_tool_funcs.append(tool_func)
```
kramaranya: These seem to be already in the trainer's tools list, so this loop would register them twice

kubeflow_mcp/trainer/__init__.py
```diff
@@ -46,108 +57,364 @@
...
+    "wait_for_training": "Block until job reaches target status (Complete/Failed). Supports optional namespace.",
+    "delete_training_job": "[DESTRUCTIVE] Delete a training job permanently. Set confirmed=True to execute.",
+    "update_training_job": "Suspend or resume a training job. Pass action='suspend' or 'resume'.",
+    "inspect_crd": "List Trainer CRDs or get details for a specific one. Pass name= for details.",
+    "inspect_controller": "Inspect controller pod. Pass view='logs' or 'events'. Auto-discovers namespace.",
+    "patch_runtime": "Strategic merge patch on a ClusterTrainingRuntime. Set confirmed=True to apply.",
+    "create_runtime": "Create a new ClusterTrainingRuntime. Set confirmed=True to create.",
+    "delete_runtime": "[DESTRUCTIVE] Delete a ClusterTrainingRuntime. Lists dependent jobs first. Set confirmed=True.",
+    "health_check": "Check server health and K8s connectivity.",
+    "get_server_logs": "Get recent server logs for debugging. Filter by level.",
```
kramaranya: These are server-level tools, I dont think they should live inside

kubeflow_mcp/core/config.py
```diff
@@ -0,0 +1,309 @@
...
+
+def set_effective_persona(persona: str) -> None:
+    """Store the resolved persona at server startup."""
+    global _effective_persona
+    _effective_persona = persona
+
+
+def get_effective_persona() -> str:
+    """Return the persona resolved at server startup."""
+    return _effective_persona
```
kramaranya: This is runtime state, not config, shall this live in policy.py alongside the persona definitions?

kubeflow_mcp/core/policy.py
```diff
@@ -0,0 +1,274 @@
...
+            "run_container_training",
+            "update_training_job",
+            "inspect_crd",
+            "inspect_controller",
+        ],
+    },
+    "platform-admin": {"tools": "*"},
+}
+
+TOOL_CATEGORIES: dict[str, list[str]] = TOOL_PHASES
```
kramaranya: nit: TOOL_CATEGORIES = TOOL_PHASES is just an alias that could be the import directly

abhijeet-dhumal:
Hey @kramaranya, thanks again for the thorough pass. Addressed several of your threaded points..
The PR is now ready for the next round 😄

andreyvelich (commented):
Thanks @abhijeet-dhumal!
/lgtm
/approve

/hold @kramaranya in case you have more suggestions

kramaranya (approved):
Thank you, @abhijeet-dhumal!
LGTM
/unhold

PR #20, opened 2026-05-15
fix: do not wrap training script if it already defines def train()

### LINKED ISSUE #19: run_custom_training silently skips training when script defines def train() without calling it
### Description

### Steps to Reproduce

```
run_custom_training(
    script="""
def train():
    import torch
    print("Training started")
    for epoch in range(3):
        print(f"Epoch {epoch}")
""",
    gpu_per_node=0,
    confirmed=True,
)
```

Root Cause
```
for line in lines:
```

```
        print("Training started")
        for epoch in range(3):
            print(f"Epoch {epoch}")

```

### Expected Behavior

## Fix 

### Actual Behavior

_No response_

### Version

master

### Python Version

3.12

sjhddh:
Fixes #19.

When a user passes a custom training script that already defines `def train():` or `def train(parameters...):`, the `_make_train_func` helper previously wrapped the entire script in another `def train():` wrapper and just indented the user's code. This meant the inner function was defined but never called, leading to jobs completing with `exit code 0` but skipping training entirely.

This PR uses `ast.parse` to detect if a top-level `train` function is already defined in the script. If it is, we return the script verbatim. If `func_args` are provided, we also check the AST to ensure the user's `train` function accepts the parameters (or has `**kwargs`), and raise an early clear `ValueError` if it does not.

Added unit tests in `tests/unit/trainer/test_training_script_wrapper.py`.

### DIFF
--- a/kubeflow_mcp/trainer/api/training.py
+++ b/kubeflow_mcp/trainer/api/training.py
@@ -14,6 +14,7 @@
 
 """Training submission tools: fine_tune, run_custom_training, run_container_training."""
 
+import ast
 import logging
 import os
 import tempfile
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
hey @sjhddh , as discussed [here](https://github.com/kubeflow/mcp-server/issues/19), can you close this stale pr ?

PR #21, opened 2026-05-15
feat: add OpenTelemetry tracing for tool calls

### LINKED ISSUE #18: feat: add OpenTelemetry tracing to tool calls
### Description

Add OpenTelemetry (OTel) tracing to the Kubeflow MCP Server so every tool invocation produces a span with standard attributes - enabling distributed tracing in production deployments alongside the existing structured logging.

### Use Case

A platform team deploys the MCP server in-cluster and routes traffic to a Jaeger or Tempo backend. Today they can correlate logs via `correlation_id`, but have no cross-service trace propagation, no latency histograms per tool, and no way to see in a tracing UI which tool calls are slow or failing.

With OTel:
- Each `tool_call` becomes a span with `tool.name`, `tool.success`, `tool.duration_ms`, `kubeflow.persona`, and the existing `correlation_id` as the trace ID
- Spans propagate W3C TraceContext headers when the HTTP transport is used,  letting downstream Kubeflow SDK calls appear as child spans
- Operators can alert on `tool.success=false` rate or P99 latency per tool without parsing logs

### Proposed Solution

The natural hook point is `_audit_wrap` in `core/server.py` - every tool call already passes through it with `tool_name`, `duration_ms`, `correlation_id`, and `success`. OTel spans slot in without touching any tool implementations.

**Changes:**
- [ ] Add `opentelemetry-sdk`, `opentelemetry-exporter-otlp` as optional deps
- [ ] Add `core/telemetry.py` - `setup_tracing(endpoint, service_name)` and
      `get_tracer()` helpers; no-op tracer if OTel is not installed
- [ ] Extend `_audit_wrap` to open a span per tool call, set attributes (`tool.name`, `tool.success`, `tool.duration_ms`, `kubeflow.persona`), and record exceptions
- [ ] Wire `setup_tracing()` into the CLI via  `--otel-endpoint` / `KUBEFLOW_MCP_OTEL_ENDPOINT` env var
- [ ] Add unit tests for `core/telemetry.py` (no-op path + span attribute assertions)
- [ ] Document the feature in `README.md` under a new Observability section

**Out of scope for this github issue(follow-up):**
- Metrics (counters/histograms) 
- Trace propagation into Kubeflow SDK calls

### Component

Core

priyank766:
## Description
 
 Adds optional OpenTelemetry tracing for tool calls in Kubeflow MCP Server.
 
 - Adds `core/telemetry.py` with `setup_tracing()` and `get_tracer()` plus safe no-op fallback when OTel deps are unavailable.
 - Instruments `core.server._audit_wrap` to create one span per tool invocation, set tool/persona/duration/success attributes, attach `correlation_id`, and record
 exceptions.
 - Wires tracing config through CLI/env/config:
   - `--otel-endpoint`
   - `KUBEFLOW_MCP_OTEL_ENDPOINT`
   - `observability.otel_endpoint`
 - Adds unit tests for no-op path, provider setup/reuse, endpoint validation, and span behavior.
 - Updates README with an Observability section.
 
 Important compatibility note: `correlation_id` semantics are preserved and exposed as a span attribute; it is not remapped to OTel trace ID.
 
 ## Type of Change
 
 - [x] feat: New feature
 
 ## Checklist
 
 ## Related Issues
 
 Fixes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,14 +215,17 @@ OpenTelemetry tracing is optional and can be enabled without changing tool code.
 - Enable tracing with CLI flag or env var:
 
 ```bash
-kubeflow-mcp serve --otel-endpoint http://localhost:4318/v1/traces
+kubeflow-mcp serve --otel-endpoint http://localhost:4318
 # or
-export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318/v1/traces
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
 kubeflow-mcp serve
 ```
 
 Each tool invocation emits a span with attributes:
-`tool.name`, `tool.success`, `tool.duration_ms`, `kubeflow.persona`, and `correlation_id`.
+`tool.name`, `tool.args_preview`, `tool.success`, `tool.duration_ms`, `kubeflow.persona`, and `correlation_id`.
+
+> **Note:** `kubeflow-mcp agent --otel-endpoint ...` emits spans under a separate
+> `kubeflow-mcp-agent` service in Jaeger, distinct from the `kubeflow-mcp` server spans.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ OpenTelemetry tracing is optional and can be enabled without changing tool code.
 ```bash
 kubeflow-mcp serve --otel-endpoint http://localhost:4318/v1/traces
 # or
-export KUBEFLOW_MCP_OTEL_ENDPOINT=http://localhost:4318/v1/traces
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318/v1/traces
 kubeflow-mcp serve
 ```
 
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ kubeflow-mcp serve \
   --instruction-tier full \       # full | compact | minimal
   --transport stdio \             # stdio | http | sse
   --auth-token SECRET \           # bearer token for HTTP auth (dev/staging)
+  --otel-endpoint URL \           # OTLP HTTP endpoint (optional tracing)
   --log-level INFO \              # DEBUG | INFO | WARNING | ERROR
   --log-format console \          # console | json (auto-detected if omitted)
   --no-banner                     # suppress startup banner
@@ -205,6 +206,23 @@ kubeflow-mcp agent \
 ```
 
 </details>
+
+## Observability
+
+OpenTelemetry tracing is optional and can be enabled without changing tool code.
+
+- Install optional dependencies: `pip install ".[otel]"`
+- Enable tracing with CLI flag or env var:
+
+```bash
+kubeflow-mcp serve --otel-endpoint http://localhost:4318/v1/traces
+# or
+export KUBEFLOW_MCP_OTEL_ENDPOINT=http://localhost:4318/v1/traces
+kubeflow-mcp serve
+```
+
+Each tool invocation emits a span with attributes:
+`tool.name`, `tool.success`, `tool.duration_ms`, `kubeflow.persona`, and `correlation_id`.
 
 ## Development
 
--- a/kubeflow_mcp/cli.py
+++ b/kubeflow_mcp/cli.py
@@ -98,7 +98,7 @@ def cli() -> None:
     "--otel-endpoint",
     default=None,
     help="OpenTelemetry OTLP HTTP endpoint for tracing. "
-    "Falls back to KUBEFLOW_MCP_OTEL_ENDPOINT env var, config file.",
+    "Falls back to OTEL_EXPORTER_OTLP_ENDPOINT env var, config file.",
 )
 def serve(
     clients: str | None,
--- a/kubeflow_mcp/cli.py
+++ b/kubeflow_mcp/cli.py
@@ -94,6 +94,12 @@ def cli() -> None:
     "Falls back to KUBEFLOW_MCP_AUTH_TOKEN env var, config file. "
     "Ignored for stdio transport.",
 )
+@click.option(
+    "--otel-endpoint",
+    default=None,
+    help="OpenTelemetry OTLP HTTP endpoint for tracing. "
+    "Falls back to KUBEFLOW_MCP_OTEL_ENDPOINT env var, config file.",
+)
 def serve(
     clients: str | None,
     persona: str | None,
@@ -104,6 +110,7 @@ def serve(
     instruction_tier: str | None,
     no_banner: bool,
     auth_token: str | None,
+    otel_endpoint: str | None,
 ) -> None:
     """Start the MCP server.
 
@@ -116,6 +123,7 @@ def serve(
     from kubeflow_mcp.core.logging import setup_logging
     from kubeflow_mcp.core.resilience import configure_circuit_breaker
     from kubeflow_mcp.core.server import configure_resilience, create_server
+    from kubeflow_mcp.core.telemetry import setup_tracing
 
     cfg = load_config()
 
@@ -128,8 +136,11 @@ def serve(
 
     if auth_token:
         cfg.auth.auth_token = auth_token
+    if otel_endpoint:
+        cfg.observability.otel_endpoint = otel_endpoint
 
     logger = setup_logging(level=log_level, format=log_format)
+    tracing_enabled = setup_tracing(endpoint=cfg.observability.otel_endpoint)
     logger.info(
         "Starting kubeflow-mcp",
         extra={
@@ -138,6 +149,7 @@ def serve(
             "transport": transport,
             "mode": mode,
             "instruction_tier": instruction_tier,
+            "tracing_enabled": tracing_enabled,
         },
     )
 
--- a/kubeflow_mcp/core/config.py
+++ b/kubeflow_mcp/core/config.py
@@ -51,7 +51,7 @@ Example config file (~/.kubeflow-mcp.yaml):
             format: json
 
         observability:
-            otel_endpoint: http://localhost:4318/v1/traces
+            otel_endpoint: http://localhost:4318
 
 Namespace restrictions are enforced via ``~/.kf-mcp-policy.yaml``
 (``policy.namespaces``), not through server config.
--- a/kubeflow_mcp/core/config.py
+++ b/kubeflow_mcp/core/config.py
@@ -242,7 +242,7 @@ def load_config(config_path: Path | None = None) -> Config:
     observability_file = file_config.get("observability", {})
     observability = ObservabilityConfig(
         otel_endpoint=os.getenv(
-            "KUBEFLOW_MCP_OTEL_ENDPOINT",
+            "OTEL_EXPORTER_OTLP_ENDPOINT",
             observability_file.get("otel_endpoint"),
         )
     )
--- a/kubeflow_mcp/core/config.py
+++ b/kubeflow_mcp/core/config.py
@@ -50,6 +50,9 @@ Example config file (~/.kubeflow-mcp.yaml):
             level: INFO
             format: json
 
+        observability:
+            otel_endpoint: http://localhost:4318/v1/traces
+
 Namespace restrictions are enforced via ``~/.kf-mcp-policy.yaml``
 (``policy.namespaces``), not through server config.
 """
@@ -135,6 +138,12 @@ class LoggingConfig(BaseModel):
     format: str | None = Field(default=None)
 
 
+class ObservabilityConfig(BaseModel):
+    """Observability configuration."""
+
+    otel_endpoint: str | None = Field(default=None)
+
+
 class Config(BaseModel):
     """Root configuration."""
 
@@ -144,6 +153,7 @@ class Config(BaseModel):
     trainer: TrainerConfig = Field(default_factory=TrainerConfig)
     optimizer: OptimizerConfig = Field(default_factory=OptimizerConfig)
     logging: LoggingConfig = Field(default_factory=LoggingConfig)
+    observability: ObservabilityConfig = Field(default_factory=ObservabilityConfig)
 
 
 def _find_config_file() -> Path | None:
@@ -229,6 +239,14 @@ def load_config(config_path: Path | None = None) -> Config:
         format=os.getenv("LOG_FORMAT", logging_file.get("format")),
     )
 
+    observability_file = file_config.get("observability", {})
+    observability = ObservabilityConfig(
+        otel_endpoint=os.getenv(
+            "KUBEFLOW_MCP_OTEL_ENDPOINT",
+            observability_file.get("otel_endpoint"),
+        )
+    )
+
     # Build client-specific configs
     trainer_file = file_config.get("trainer", {})
     trainer = TrainerConfig(
@@ -283,6 +301,7 @@ def load_config(config_path: Path | None = None) -> Config:
         trainer=trainer,
         optimizer=optimizer,
         logging=logging_config,
+        observability=observability,
     )
 
 
--- a/kubeflow_mcp/core/server.py
+++ b/kubeflow_mcp/core/server.py
@@ -21,6 +21,7 @@ Designed for extensibility:
 
 import functools
 import importlib
+import json
 import logging
 import re
 import time
@@ -118,6 +119,11 @@ def _audit_wrap(tool_func):
             span.set_attribute("tool.name", tool_name)
             span.set_attribute("kubeflow.persona", persona)
             span.set_attribute("correlation_id", cid)
+            masked = mask_sensitive_data(kwargs) if kwargs else {}
+            span.set_attribute(
+                "tool.args_preview",
+                json.dumps(masked, default=str)[:300],
+            )
 
             if _rate_limiter is not None and not _rate_limiter.acquire():
                 duration_ms = int((time.monotonic() - start) * 1000)
@@ -140,7 +146,6 @@ def _audit_wrap(tool_func):
                     "error_code": ErrorCode.CIRCUIT_OPEN,
                 }
 
-            masked = mask_sensitive_data(kwargs) if kwargs else {}
             try:
                 result = tool_func(**kwargs)
                 duration_ms = int((time.monotonic() - start) * 1000)
--- a/kubeflow_mcp/core/server.py
+++ b/kubeflow_mcp/core/server.py
@@ -27,8 +27,7 @@ import re
 import time
 from typing import Any
 
-from fastmcp import Context, FastMCP
-from fastmcp.dependencies import CurrentContext
+from fastmcp import FastMCP
 
 from kubeflow_mcp.common.constants import (
     TOOL_NEXT_HINTS,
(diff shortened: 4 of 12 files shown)


abhijeet-dhumal:
Hey @priyank766 , this looks great 🚀 
Thanks for working on this !

pyproject.toml
```diff
@@ -40,6 +40,10 @@ docs = [
     "sphinx-copybutton>=0.5",
     "sphinx-design>=0.5",
 ]
+otel = [
+    "opentelemetry-exporter-otlp>=1.25.0",
```
abhijeet-dhumal:
```suggestion
    "opentelemetry-exporter-otlp-proto-http>=1.25.0",
```
this avoids installation of unnecessary GRPC subpackage

kubeflow_mcp/core/server.py
```diff
@@ -88,66 +94,84 @@ def _audit_wrap(tool_func):
...
-                    "correlation_id": cid,
-                    "tool": tool_name,
-                    "parameters": masked,
-                    "success": False,
-                    "duration_ms": duration_ms,
-                },
-                exc_info=True,
-            )
-            raise
+        with tracer.start_as_current_span("tool_call") as span:
```
abhijeet-dhumal: I tried it just now and it seems span name "tool_call" makes it hard to filter by tool in tracing UIs, Maybe we can consider naming it after the tool: f"tool:{tool_name}" or even just tool_name. The attribute tool.name is still good to keep for structured querying, but the name gives context at a glance.
wdyt?

priyank766: Thanks for Reviewing @abhijeet-dhumal 
I will push both changes and then you can review it later whenever you get time 
Yes I think this naming scheme is much better I will change it .. 👍🏻

pyproject.toml
priyank766: Thanks for the Suggestion 
I will change the package as well

kubeflow_mcp/core/server.py
```diff
@@ -88,66 +94,84 @@ def _audit_wrap(tool_func):
...
+                        "duration_ms": duration_ms,
+                    },
+                )
+                return _inject_meta(result, tool_name)
+            except Exception as exc:
+                duration_ms = int((time.monotonic() - start) * 1000)
+                breaker.record_failure()
+                span.set_attribute("tool.success", False)
+                span.set_attribute("tool.duration_ms", duration_ms)
+                span.record_exception(exc)
```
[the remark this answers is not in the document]
abhijeet-dhumal: @priyank766  can we add span.set_status(StatusCode.ERROR) on exception here ?
span.set_status(Status(StatusCode.ERROR, str(exc)))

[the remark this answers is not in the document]
abhijeet-dhumal: this will make sure to spot failures easily in the trace list, wdyt?

kubeflow_mcp/core/server.py
```diff
@@ -88,66 +94,84 @@ def _audit_wrap(tool_func):
...
-        if not breaker.can_execute():
-            logger.warning("circuit_open", extra={"tool": tool_name})
-            return {
-                "error": f"Circuit breaker open for '{tool_name}' — K8s API may be degraded. Retries automatically after recovery timeout.",
-                "error_code": ErrorCode.CIRCUIT_OPEN,
-            }
-
         cid = with_correlation_id()
-        masked = mask_sensitive_data(kwargs) if kwargs else {}
+        tracer = get_tracer("kubeflow_mcp.tools")
```
abhijeet-dhumal: @priyank766 can you move this outside wrapper closure? 
get_tracer is cheap but calling it per-invocation seems semantically wrong, wdyt?

kubeflow_mcp/core/server.py
```diff
@@ -88,66 +94,84 @@ def _audit_wrap(tool_func):
...
-                    "correlation_id": cid,
-                    "tool": tool_name,
-                    "parameters": masked,
-                    "success": False,
-                    "duration_ms": duration_ms,
-                },
-                exc_info=True,
-            )
-            raise
+        with tracer.start_as_current_span(f"tool:{tool_name}") as span:
```
abhijeet-dhumal: @priyank766  thinking can we use SpanKind.CLIENT for tool spans here?
Tool calls invoke an external service i.e. K8s API. SpanKind.CLIENT is semantically correct and enables better Jaeger dependency graph rendering..

kubeflow_mcp/core/telemetry.py
```diff
@@ -0,0 +1,125 @@
...
+
+class _NoopSpan:
+    def set_attribute(self, key: str, value: Any) -> None:
+        return None
+
+    def record_exception(self, exception: BaseException) -> None:
+        return None
+
+
+class _NoopTracer:
```
abhijeet-dhumal, line 53: _NoopTracer should accept **kwargs here
Future callers passing kind=SpanKind.CLIENT will otherwise get a TypeError in no-op mode.

priyank766:
Thanks for the review @abhijeet-dhumal  All suggestions have been addressed. Please take another look when you get a chance.

README.md
```diff
@@ -165,6 +166,23 @@ kubeflow-mcp agent \
...
+
+OpenTelemetry tracing is optional and can be enabled without changing tool code.
+
+- Install optional dependencies: `pip install ".[otel]"`
+- Enable tracing with CLI flag or env var:
+
+```bash
+kubeflow-mcp serve --otel-endpoint http://localhost:4318/v1/traces
+# or
+export KUBEFLOW_MCP_OTEL_ENDPOINT=http://localhost:4318/v1/traces
```
abhijeet-dhumal:
```suggestion
export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318/v1/traces
```

kubeflow_mcp/core/config.py
```diff
@@ -229,6 +239,14 @@ def load_config(config_path: Path | None = None) -> Config:
         format=os.getenv("LOG_FORMAT", logging_file.get("format")),
     )
 
+    observability_file = file_config.get("observability", {})
+    observability = ObservabilityConfig(
+        otel_endpoint=os.getenv(
+            "KUBEFLOW_MCP_OTEL_ENDPOINT",
```
abhijeet-dhumal:
```suggestion
            "OTEL_EXPORTER_OTLP_ENDPOINT",
```

kubeflow_mcp/cli.py
```diff
@@ -94,6 +94,12 @@ def cli() -> None:
     "Falls back to KUBEFLOW_MCP_AUTH_TOKEN env var, config file. "
     "Ignored for stdio transport.",
 )
+@click.option(
+    "--otel-endpoint",
+    default=None,
+    help="OpenTelemetry OTLP HTTP endpoint for tracing. "
+    "Falls back to KUBEFLOW_MCP_OTEL_ENDPOINT env var, config file.",
```
abhijeet-dhumal:
```suggestion
    "Falls back to OTEL_EXPORTER_OTLP_ENDPOINT env var, config file.",
```

kubeflow_mcp/core/server.py
```diff
@@ -84,70 +99,95 @@ def _inject_meta(result: Any, tool_name: str) -> Any:
...
-            )
-            raise
+        span_kwargs: dict[str, Any] = {}
+        if SpanKind is not None:
+            span_kwargs["kind"] = SpanKind.CLIENT
+
+        with tracer.start_as_current_span(
+            f"tool:{tool_name}", **span_kwargs
+        ) as span:
+            span.set_attribute("tool.name", tool_name)
```
abhijeet-dhumal: Minor nit: would be great to also surface user.id and mcp.session_id on these spans for per-session/per-user filtering in Jaeger. That will need identity propagation middleware (ContextVars populated from the MCP request context) as a prerequisite..

abhijeet-dhumal: happy to follow up with a separate PR for that once this lands. Marking as a non-blocking suggestion.

priyank766: Makes sense, having `user.id` and `session_id` on the spans would definitely help with debugging. Happy to take a crack at it once this lands, just let me know.

abhijeet-dhumal:
Hey @priyank766 , I tried this out locally against OTel collector stack, works cleanly end-to-end. 
Spans show up in Jaeger with the right service name, tool: prefix naming is much better for filtering. 
Thanks you for being consistent and addressing all the previous feedback so promptly. 🚀 🙌 

putting few nits otherwise lgtm..

kubeflow_mcp/core/telemetry.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if hasattr(current_provider, "add_span_processor"):
+            current_provider.add_span_processor(processor)
+        else:
+            resource = Resource.create({"service.name": service_name})
+            provider = TracerProvider(resource=resource)
+            provider.add_span_processor(processor)
+            _otel_trace.set_tracer_provider(provider)
+
+        _tracing_initialized = True
```
abhijeet-dhumal, line 132: Missing atexit.register(provider.shutdown) here. Without it, in-flight spans in the BatchSpanProcessor queue get silently dropped on server shutdown, and the background thread can deadlock if the collector is unreachable. One line:
```
import atexit
atexit.register(provider.shutdown)
```

kubeflow_mcp/core/server.py
abhijeet-dhumal: Can we also add tool.args_preview here? Without it you can't reconstruct what parameters caused a failure from Jaeger alone.. you'd have to cross-reference the audit log. Something like:
```
import json
span.set_attribute("tool.args_preview", json.dumps(mask_sensitive_data(kwargs), default=str)[:300])
```

kubeflow_mcp/core/server.py
```diff
@@ -84,70 +99,95 @@ def _inject_meta(result: Any, tool_name: str) -> Any:
...
+                    "error_code": ErrorCode.RATE_LIMITED,
+                }
+
+            breaker = get_breaker(tool_name)
+            if not breaker.can_execute():
+                duration_ms = int((time.monotonic() - start) * 1000)
+                span.set_attribute("tool.success", False)
+                span.set_attribute("tool.duration_ms", duration_ms)
+                logger.warning("circuit_open", extra={"tool": tool_name})
+                return {
```
abhijeet-dhumal, line 168: This path sets tool.success=False on the span but there's no test covering it. 
Worth adding a test with _FakeBreaker(can_execute=False) asserting the attributes are set before the early return.

README.md
```diff
@@ -165,6 +166,23 @@ kubeflow-mcp agent \
...
 
+## Observability
+
+OpenTelemetry tracing is optional and can be enabled without changing tool code.
+
+- Install optional dependencies: `pip install ".[otel]"`
+- Enable tracing with CLI flag or env var:
+
+```bash
+kubeflow-mcp serve --otel-endpoint http://localhost:4318/v1/traces
```
abhijeet-dhumal: Heads up: setup_tracing() auto-appends /v1/traces to whatever is passed, so the example in the README produces http://localhost:4318/v1/traces/v1/traces — double path. Either change the README example to the base URL (http://localhost:4318) or stop auto-appending in code and accept the full path as-is. The latter is less surprising.

README.md
```diff
@@ -165,6 +166,23 @@ kubeflow-mcp agent \
...
+# or
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318/v1/traces
+kubeflow-mcp serve
+```
+
+Each tool invocation emits a span with attributes:
+`tool.name`, `tool.success`, `tool.duration_ms`, `kubeflow.persona`, and `correlation_id`.
+
```
abhijeet-dhumal, line 229: Worth a one-liner noting that kubeflow-mcp agent  --otel-endpoint ... emits a separate kubeflow-mcp-agent service in Jaeger. Without this, users running the agent will wonder why they only see server-side spans and think tracing is broken.

priyank766:
I've addressed all the nits in the latest commit. Let me know if everything looks good!! 
Thanks for testing it out and for the kind words ! 🚀
@abhijeet-dhumal

abhijeet-dhumal:
Hey @priyank766 👋
Apologies for raising this late in the review cycle.. stumbled across something that's worth addressing before this lands. 
There's a dedicated MCP-specific semantic convention that shipped in OTel v1.39: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/.  Would it be good to address it now ?
It defines exactly how MCP server tool call spans should be instrumented, and i afraid our current attribute names don't align.
WDYT? @andreyvelich @astefanutti @kramaranya

priyank766:
Hey @abhijeet-dhumal No worries at all, and good call flagging this  definitely better to align now than retrofit later.

I went through the [MCP semconv spec](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/) and yeah, our attribute names are off. Here's the gap:

| Aspect | What we have now | What the spec expects |
|---|---|---|
| Span name | `tool:{name}` | `tools/call {name}` |
| SpanKind | `CLIENT` | `SERVER` (we're the MCP server receiving the call) |
| Tool name attribute | `tool.name` | `gen_ai.tool.name` |
| Method attribute | _(missing)_ | `mcp.method.name = "tools/call"` |
| Operation attribute | _(missing)_ | `gen_ai.operation.name = "execute_tool"` |
| Session & Request IDs | _(missing)_ | `mcp.session.id`, `mcp.request.id` |
| Protocol version | _(missing)_ | `mcp.protocol.version` |

One Good thing is FastMCP 3.2.4 already exposes `ctx.session_id` and `ctx.request_id` through its `Context` object, so pulling in the MCP-specific attributes should be straightforward.

I'd lean towards doing the full alignment in one go since the scope is manageable - mostly attribute renames in `_audit_wrap`, SpanKind change, and wiring in `Context` for session/request IDs. Our custom attributes (`tool.success`, `tool.duration_ms`, `kubeflow.persona` etc.) would stay as additional enrichment on top. Tests will be updated to match the new attribute names.

One thing worth noting  these conventions are still in **Development** status and there are no Python package constants for them yet, so we'd go with string literals for now.

Can push a follow-up commit on this PR . Let me know if that what we decide or if you'd rather keep it scoped differently.

abhijeet-dhumal:
Hey @priyank766, great investigation, fully agree with the alignment plan, go ahead with the commit on this PR 👍
Also, still worth bumping >=1.25.0 to >=1.29.0 in pyproject.toml while you're in there.

priyank766:
I commited the full alignment & switched the spans to SERVER, renamed them to  `tools/call{name}`  , and wired in the FastMCP Context so we're grabbing the actual  `session_id`  and  `request_id`  at runtime. Also added the protocol version from the SDK and set error.type  on exceptions.       
                                                                                                                                                             
I also bumped the `OTel` packages to  `>=1.29.0`  in  pyproject.toml  and updated/added tests to cover it all. Everything is passing locally.                  
@abhijeet-dhumal 
Let me know what you think!!

kubeflow_mcp/core/telemetry.py
```diff
@@ -0,0 +1,133 @@
...
+            "for example: http://localhost:4318"
+        )
+    # Append the standard OTLP trace path when only a base URL is provided,
+    # matching the OTel SDK convention for OTEL_EXPORTER_OTLP_ENDPOINT.
+    if not parsed.path or parsed.path == "/":
+        normalized_endpoint = normalized_endpoint.rstrip("/") + "/v1/traces"
+    return normalized_endpoint
+
+
+def setup_tracing(endpoint: str | None, service_name: str = "kubeflow-mcp") -> bool:
```
abhijeet-dhumal: Rename service from kubeflow-mcp to kubeflow-mcp-server so Jaeger clearly separates server vs agent (kubeflow-mcp-agent) #4

kubeflow_mcp/core/telemetry.py
```diff
@@ -0,0 +1,133 @@
...
+        return False
+
+    with _setup_lock:
+        if _tracing_initialized:
+            if _configured_endpoint and _configured_endpoint != normalized_endpoint:
+                logger.warning(
+                    "Tracing already initialized for endpoint '%s'; ignoring new endpoint '%s'.",
+                    _configured_endpoint,
+                    normalized_endpoint,
+                )
```
abhijeet-dhumal, line 113: Add export timeouts.. without these, unreachable collector hangs shutdown
export_timeout_millis=2000 on processor + exporter

kubeflow_mcp/core/telemetry.py
abhijeet-dhumal: Fall back to env var inside setup_tracing, not only in config loader
setup_tracing() can read OTEL_EXPORTER_OTLP_ENDPOINT when endpoint=None

kubeflow_mcp/core/server.py
```diff
@@ -84,70 +109,119 @@ def _inject_meta(result: Any, tool_name: str) -> Any:
 
 def _audit_wrap(tool_func):
     """Wrap a tool function with rate limiting, circuit breaking, audit logging, and response metadata."""
+    tracer = get_tracer("kubeflow_mcp.tools")
 
     @functools.wraps(tool_func)
-    def wrapper(**kwargs):
+    def wrapper(ctx: Context | None = _MCP_CTX_DEFAULT, **kwargs):
```
abhijeet-dhumal: CurrentContext() DI on sync wrapper may not inject reliably..  prefer AuditIdentityMiddleware + ContextVars for user.id / mcp.session.id fallback

kubeflow_mcp/core/server.py
```diff
@@ -84,70 +109,119 @@ def _inject_meta(result: Any, tool_name: str) -> Any:
...
+                except Exception:
+                    pass
+                try:
+                    if ctx.request_id is not None:
+                        span.set_attribute("mcp.request.id", str(ctx.request_id))
+                except Exception:
+                    pass
+
+            # Custom Kubeflow enrichment
+            span.set_attribute("kubeflow.persona", persona)
```
abhijeet-dhumal, line 144: Add user.id from middleware identity bridge

abhijeet-dhumal: we can add Add middleware (core/middleware.py - AuditIdentityMiddleware) to bridge FastMCP async context to sync _audit_wrap

abhijeet-dhumal:
@priyank766 Maybe you also need to re-sign commits to fix DCO pr check

abhijeet-dhumal:
Thanks @priyank766  for the quick turnaround. It looks great !
LGTM .. just need to resolve pr checks : Can you Rebase on latest main and run ruff format to pass pr checks ?

abhijeet-dhumal:
This is amazing work 🚀 
Thanks a lot @priyank766 for your patience!!
/lgtm

abhijeet-dhumal:
/lgtm
Thanks @priyank766, I have tested this and works perfectly as expected !

abhijeet-dhumal:
@priyank766 would it be good if you raise 1 more pr resolving these 2 failing PR checks?
just need to update action references

priyank766:
Added the PR to solve the failing checks :  https://github.com/kubeflow/mcp-server/pull/49
@abhijeet-dhumal

abhijeet-dhumal:
/lgtm
/approve

OTel tracing for Phase 2 observability. Tests pass; security-audit failure is pre-existing pip-audit issue (addressed by #30).

priyank766:
resolved the conflicts

abhijeet-dhumal:
/lgtm
/approve
Thanks @priyank766 for your contribution, this is great work 🚀

PR #22, opened 2026-05-16
fix: detect user-defined train() to prevent silent training skips

### LINKED ISSUE #19: run_custom_training silently skips training when script defines def train() without calling it

algojogacor:
## Summary

Fixes #19 — `run_custom_training` silently skips training when the script defines a top-level `def train()` without calling it.

## Root Cause

`_make_train_func` unconditionally wraps the entire script inside `def train():`. When the user's script already defines `def train()` at the top level, the user's function gets nested inside the wrapper:

```python
def train():        # outer wrapper (called by SDK)
    def train():    # user's function (defined, never called)
        print("Training started")
train()             # appended by SDK
```

## Solution

Use `ast.parse` to detect whether the script already has a top-level `def train()`:
- If yes: write the script as-is (no wrapping), compile, and return the user's function directly
- If func_args provided: validate that the user's `train()` signature includes all required parameters, raising a clear `ValueError` on mismatch
- If no: use the existing wrapping behavior (backward compatible)

Edge cases handled:
- Nested `def train(self)` inside a class is not detected as top-level
- SyntaxError in script propagates correctly from `compile()`
- Module-level imports and non-train code in user scripts execute normally

## Testing

Added 11 unit tests covering:
- User-defined `train()` is detected and called directly
- Side effects and return values work correctly
- Module-level imports and helper code execute normally
- Nested `train()` methods not falsely detected
- `func_args` matching, defaults, and mismatched detection
- Syntax error propagation
- Existing wrapping behavior preserved for scripts without `def train()`

All 51 tests pass (40 existing + 11 new), zero regressions.

### DIFF
--- a/kubeflow_mcp/trainer/api/training.py
+++ b/kubeflow_mcp/trainer/api/training.py
@@ -102,8 +102,57 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
 
     When *func_args* is provided, the generated function signature includes matching
     keyword parameters so the trainer can pass them at runtime.
+
+    If the script already defines a top-level ``def train()``, that function is
+    used directly — avoiding silent training skips from double-wrapping.
     """
+    import ast
+
     func_name = "train"
+
+    # Check if the script already defines a top-level train() function.
+    try:
+        tree = ast.parse(script)
+    except SyntaxError:
+        tree = None
+
+    user_train = None
+    if tree is not None:
+        for node in ast.iter_child_nodes(tree):
+            if isinstance(node, ast.FunctionDef) and node.name == func_name:
+                user_train = node
+                break
+
+    if user_train is not None:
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
hey @algojogacor , as discussed here, can you close this stale pr ? : https://github.com/kubeflow/mcp-server/issues/19

PR #23, opened 2026-05-17
Fix custom training scripts that define train()

Kaladin-Stormborn:
# PR Draft: kubeflow/mcp-server#19

## Title

Fix custom training scripts that define an uncalled `train()`

## Summary

This fixes `run_custom_training` script wrapping when a user provides a script that already defines a top-level `train()` function but does not call it.

Previously `_make_train_func()` always wrapped the submitted script inside another generated `def train():`. For scripts like:

```python
def train():
    print("Training started")
```

the user-defined `train()` became a nested function and was never invoked when the generated wrapper ran.

The patch detects a top-level `def train()` with no module-level `train()` call and appends a call inside the generated wrapper. Scripts that already call `train()` are left alone to avoid double execution. Matching `func_args` are forwarded to a user-defined `train(...)` signature.

## Tests

```bash
uv sync --extra dev
.venv/bin/ruff check kubeflow_mcp/trainer/api/training.py kubeflow_mcp/trainer/api/sdk_contracts_test.py
.venv/bin/python -m pytest kubeflow_mcp/trainer/api/sdk_contracts_test.py -q
.venv/bin/python -m pytest kubeflow_mcp/trainer/api/sdk_contracts_test.py tests/unit/trainer/test_architecture.py -q
.venv/bin/python -m pytest -q
```

Results:

- Ruff: passed.
- SDK contract test: 74 passed.
- Full local suite: 139 passed, 3 upstream dependency deprecation warnings.

Current diff size:

- Implementation: `59` insertions, `1` deletion.
- Tests: `84` insertions.
- Total: `143` insertions, `1` deletion.

## Changed Files

- `kubeflow_mcp/trainer/api/training.py`
- `kubeflow_mcp/trainer/api/sdk_contracts_test.py`

### DIFF
--- a/kubeflow_mcp/trainer/api/training.py
+++ b/kubeflow_mcp/trainer/api/training.py
@@ -14,6 +14,7 @@
 
 """Training submission tools: fine_tune, run_custom_training, run_container_training."""
 
+import ast
 import logging
 import os
 import tempfile
@@ -104,7 +105,8 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
     keyword parameters so the trainer can pass them at runtime.
     """
     func_name = "train"
-    lines = script.strip().splitlines()
+    stripped_script = script.strip()
+    lines = stripped_script.splitlines()
 
     if func_args:
         params = ", ".join(f"{k}=None" for k in func_args)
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
Hey @Kaladin-Stormborn, can you sign off your commit here ⬆️ 
`git commit -s `

abhijeet-dhumal:
Thanks @Kaladin-Stormborn for this fix 🚀 
Just adding some nit picks

kubeflow_mcp/trainer/api/training.py
```diff
@@ -125,6 +130,59 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
     return ns[func_name]
 
 
+def _uncalled_train_call(script: str, func_args: dict[str, Any] | None = None) -> str | None:
```
abhijeet-dhumal, line 133: Nit (non-blocking): thinking can we consider treating top-level async def train() like def train() for the append path (ast.AsyncFunctionDef). If we append a call, it likely needs await train(...) inside the generated sync wrapper (or document that only sync def train() is supported for now).

abhijeet-dhumal:
Hey @Kaladin-Stormborn, are you still working on this ?

abhijeet-dhumal:
The issue has been fixed and merged !
/close

Issue #10, opened 2026-05-04
Benchmark suite for trainer MCP tools


abhijeet-dhumal:
### Description

Performance benchmarks for all trainer MCP tools, split from the core server PR to keep reviews focused.

### Use Case

- Establishes baseline performance metrics before optimization work
- Enables benchmark-driven decisions for Phase 2 (OTel overhead, mode comparison)

### Proposed Solution

**Benchmarks (4 files)**
- Latency — server init, tool loading, schema generation, preview, security validation (P50/P95/P99)
- Token usage — per mode (`full`/`progressive`/`semantic`) and per tool category
- CPU profiling — hot path identification
- Memory profiling — baseline and peak usage
- Dashboard — HTML report generation (`make benchmark`)

### References
- Depends on core server (issue #7) — implementation + SDK contract tests

### Component

Core

haroon0x:
@abhijeet-dhumal  i have added the benchmark for only latency tests. now `make benchmark` generates the index.html with the latency report.
- also for fine_tune, is patching GPU pre-check acceptable in benchmark code so it runs without a cluster?
- Should benchmark output include environment metadata like Python version, OS, CPU count, commit SHA, package version?
- also should generated benchmark-results/ be ignored and overwritten locally, or do you want one committed baseline/sample report for review?
I would like to add the benchmarks of token usage , cpu profiling , mem profiling in that pr. 
can you review the current pr and lmk.

Issue #5, opened 2026-04-23
SparkClient module for Kubeflow MCP Server


tariq-hasan:
### Description

A `spark` client module for the Kubeflow MCP Server that enables AI agents to manage Spark Connect sessions and Spark batch jobs on Kubernetes through natural language, following the same modular architecture used by the `trainer` module.

This covers two complementary surfaces:
1. MCP Tools - wrapping `SparkClient` operations for session lifecycle, batch job management, and monitoring
2. Agent Skills - providing structured context (SKILL.md) that guides AI agents in helping users write, configure, debug, and deploy SparkClient code

### Component

None

abhijeet-dhumal:
Hey @vedhakoushik, if you are still working on this can you add /assign label on this issue

PR #26, opened 2026-05-18
feat: add latency benchmarking and HTML reporting for trainer MCP tools

haroon0x:
First thing: this is not a vibe coded AI slop PR.

I have added the initial benchmark suite for trainer MCP tools as mentioned in #10 . This PR focuses only on latency benchmarks.

Only `tests/benchmarks/report.py`  is AI generated. The benchmark cases, pytest setup, JSON format, and benchmark flow are written in a simple way so it is easy to review and extend later.

## What is implemented

This PR adds latency benchmarks under:

```text
tests/benchmarks/test_latency.py
```

The benchmarks use `pytest-benchmark` for timing. After pytest finishes, `tests/benchmarks/conftest.py` collects the benchmark samples and writes the local result files.

Run:

```bash
uv run pytest tests/benchmarks/
```

Generated files:

```text
benchmark-results/latency.json
benchmark-results/index.html
```

`benchmark-results/` is ignored because these files are generated locally.

## Latency benchmarks added

Current latency suite covers:

- server init in `full` mode
- server init in `progressive` mode
- server init in `semantic` mode
- dynamic tool registry init
- dynamic tool discovery:
  - `list_tools`
  - `describe_tools`
  - `find_tools`
- trainer schema and metadata scan
- preview tool paths:
  - `fine_tune`
  - `run_custom_training`
  - `run_container_training`
- security validation:
  - `validate_k8s_name`
  - `validate_resource_limits`
  - `is_safe_python_code`

Preview benchmarks use `confirmed=False`, so they do not submit jobs to Kubernetes.

For `fine_tune`, the GPU pre-check is patched inside the benchmark so the benchmark can run without needing a real cluster or GPU.

## Metrics

The JSON output contains:

- P50
- P95
- P99
- min
- max

P95 and P99 are included because average or P50 alone will not show tail latency clearly.

Example output shape:

```json
{
  "suite": "latency",
  "iterations": 100,
  "warmup": 5,
  "results": [
    {
      "name": "server_init_full",
      "unit": "ms",
      "p50": 18.1,
      "p95": 24.5,
      "p99": 31.2,
      "min": 15.9,
      "max": 40.8
    }
  ]
}
```

## HTML report

The HTML report is generated at:

```text
benchmark-results/index.html
```

It shows:

- summary cards
- latency table
- P50, P95, P99
- min and max
- P99/P50 tail ratio
- visual spread bars

The report code is kept separate so future benchmark suites can also be added to the same report.

## Future plan

Next benchmark suites can be added as separate files:

```text
tests/benchmarks/test_token_usage.py
tests/benchmarks/test_cpu_profile.py
tests/benchmarks/test_memory.py
```

Rough plan:

- token usage: custom estimator because it is not a timing benchmark
- CPU profile: Python `cProfile`
- memory profile: Python `tracemalloc`

All of them can still run through pytest and write JSON files into `benchmark-results/`.

## Validation done

I ran:

```bash
uv run pytest tests/benchmarks/
uv run ruff check tests/benchmarks/test_latency.py tests/benchmarks/conftest.py tests/benchmarks/report.py
uv run ruff format --check tests/benchmarks/test_latency.py tests/benchmarks/conftest.py tests/benchmarks/report.py
uv run python -m tests.benchmarks.report
uv lock --check
```

All passed locally.

There are 3 warnings during the benchmark run. These warnings come from the installed Kubeflow dependency using old Pydantic class-based config. They are not from this benchmark code.

## Type of Change

- [x] feat: New feature
- [ ] fix: Bug fix
- [ ] revert: Revert a change

## Checklist

- [x] Benchmark tests pass locally
- [x] Linting passes for benchmark files
- [x] Documentation updated


abhijeet-dhumal:
Thanks for this @haroon0x 🚀 
scope is right for a first PR. Just few nit picks

tests/benchmarks/test_latency.py
```diff
@@ -0,0 +1,214 @@
...
+    fine_tune,
+    run_container_training,
+    run_custom_training,
+)
+
+DEFAULT_ITERATIONS = 100
+DEFAULT_WARMUP = 5
+OUTPUT_DIR = Path("benchmark-results")
+
+def measure_latency_ms(func: Callable[[], object] , iterations = DEFAULT_ITERATIONS , warmup = DEFAULT_WARMUP) -> dict[str,float]:
```
abhijeet-dhumal: thinking as we gonna introduce more benchmark types in future .. instead of hand-rolled loops can we use native pytest-benchmark utilities ? : https://pypi.org/project/pytest-benchmark/

tests/benchmarks/benchmarks_runner.py
```diff
@@ -0,0 +1,12 @@
+from tests.benchmarks.report import generate_html_report
+from tests.benchmarks.test_latency import run_latency_benchmarks
+
+
+def run_all_benchmarks():
```
abhijeet-dhumal: I think we can drop this file, can use conftest.py instead - a pytest_terminal_summary can hook run automatically with pytest tests/benchmarks/, no separate make target needed.
wdyt?

tests/benchmarks/test_latency.py
haroon0x: I was thinking of using pytest-benchmark initially then i thought since we also need P50/P95/P99 and a custom HTML report, a simple custom loop might be easier to control. But your point makes sense. Since we may add more benchmark types later, using pytest-benchmark for the latency benchmarks is cleaner than maintaining our own timing logic.

tests/benchmarks/benchmarks_runner.py
haroon0x: Yeah, makes sense if we are proceeding with pytest-benchmarks.
I’ll drop benchmarks_runner.py and move the shared setup into tests/benchmarks/conftest.py.

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import Any
+
+import pytest
+
+from tests.benchmarks.report import generate_report
+
+DEFAULT_ITERATIONS = 100
+DEFAULT_WARMUP = 5
+OUTPUT_DIR = Path("benchmark-results")
+LATENCY_RESULTS: list[dict[str, object]] = []
```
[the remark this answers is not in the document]
haroon0x: Yes, this is a valid point.

For this PR, I think we should keep the benchmark execution single-process for now. Latency benchmark results can become noisy with xdist because multiple workers will be sharing the same CPU resources. Also, the current JSON collection uses a module-level list, so running it across workers may not collect results correctly.

I’ll keep the benchmark run single-process and we can revisit xdist support later if needed.

haroon0x:
@abhijeet-dhumal I have made the changes , can you review ?

tests/benchmarks/benchmarks_utils.py
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
abhijeet-dhumal: @haroon0x Why this empty file exists ?

tests/benchmarks/conftest.py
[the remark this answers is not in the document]
abhijeet-dhumal: hey it seems a Module-level mutable list - results will accumulate if pytest runs multiple times in the same process Please reset it in pytest_configure : 
```
def pytest_configure(config):
    global LATENCY_RESULTS
    LATENCY_RESULTS = []
```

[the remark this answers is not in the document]
abhijeet-dhumal: wdyt? @copilot

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+    benchmark: Any,
+) -> Iterator[Callable[[str, Callable[[], object]], None]]:
+    def run(name: str, func: Callable[[], object]) -> None:
+        benchmark.pedantic(
+            func,
+            rounds=DEFAULT_ITERATIONS,
+            warmup_rounds=DEFAULT_WARMUP,
+            iterations=1,
+        )
+        samples_ms = [sample * 1_000 for sample in benchmark.stats.stats.data]
```
abhijeet-dhumal: benchmark.stats.stats.data is accessing internal pytest-benchmark structure. Use the stable dict API instead: benchmark.stats["data"].

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+        LATENCY_RESULTS.append(_latency_result(name, samples_ms))
+
+    yield run
+
+
+def pytest_terminal_summary(terminalreporter: Any, exitstatus: int, config: Any) -> None:
+    if not LATENCY_RESULTS:
+        return
+
+    OUTPUT_DIR.mkdir(exist_ok=True)
```
abhijeet-dhumal, line 68: Relative path assumes CWD is project root. Will silently write to the wrong place if pytest is run from a subdirectory. Pin it
```suggestion
    OUTPUT_DIR = Path(__file__).parent.parent.parent / "benchmark-results"
```

tests/benchmarks/report.py
```diff
@@ -0,0 +1,456 @@
+"""Generate a static HTML report for benchmark results."""
+
+import html
+import json
+from collections.abc import Callable
+from pathlib import Path
+from typing import Any
+
+RESULTS_DIR = Path("benchmark-results")
```
abhijeet-dhumal:
```suggestion
RESULTS_DIR = Path(__file__).parent.parent.parent / "benchmark-results"
```
same as above

tests/benchmarks/test_latency.py
```diff
@@ -0,0 +1,158 @@
...
+        ("preview_custom_training", preview_custom_training),
+        ("preview_container_training", preview_container_training),
+    ],
+)
+def test_preview_tools_latency(
+    record_latency_benchmark: Callable[[str, Callable[[], object]], None],
+    name: str,
+    benchmark_func: Callable[[], object],
+) -> None:
+    original_gpu_check = training._check_gpu_available
```
abhijeet-dhumal, line 156: If _check_gpu_available is ever renamed, the patch silently stops working and the benchmark fails with a cluster error rather than a clear test failure. Add a guard before the patch:
```
assert hasattr(training, "_check_gpu_available"), (
    "_check_gpu_available renamed — update benchmark patch"
)
```

haroon0x:
@abhijeet-dhumal I have resolved the issues as you have mentioned. Could you take a look and let me know if this is ready to be merged?

After this i was thinking on working issue #5 . Do you have any specific thoughts or req before i start working on that?

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+) -> Iterator[Callable[[str, Callable[[], object]], None]]:
+    def run(name: str, func: Callable[[], object]) -> None:
+        benchmark.pedantic(
+            func,
+            rounds=DEFAULT_ITERATIONS,
+            warmup_rounds=DEFAULT_WARMUP,
+            iterations=1,
+        )
+        samples_ms = [sample * 1_000 for sample in benchmark.stats["data"]]
+        LATENCY_RESULTS.append(_latency_result(name, samples_ms))
```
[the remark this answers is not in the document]
haroon0x: @abhijeet-dhumal Copilot is suggesting to use benchmark.stats.data

abhijeet-dhumal:
Benchmark suite is valuable but this is XL scope. Please confirm: (1) benchmark-results/ stays gitignored, (2) GPU pre-check patching is acceptable for CI-less runs, (3) split latency report from token/CPU profiling into follow-up PRs if possible.

/hold

haroon0x:
@abhijeet-dhumal yes all these are confirmed

abhijeet-dhumal:
@haroon0x can you rebase this PR?

tests/benchmarks/benchmarks_utils.py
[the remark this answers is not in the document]
abhijeet-dhumal: > [@haroon0x](https://github.com/haroon0x) Why this empty file exists ?

Did you get time to resolve this and cleanup empty file?

tests/benchmarks/benchmarks_utils.py
abhijeet-dhumal, line 1: Why this empty file is needed?

haroon0x: When we want to add more util func , we can use it. Currently there is no use. i will  remove it for now.

abhijeet-dhumal:
@haroon0x  can you check why pr checks are failing ? Have you rebased your PR on main branch ?

abhijeet-dhumal:
@haroon0x can you fix PR title for PR checks pass

haroon0x:
@abhijeet-dhumal renamed the pr title , now ready for merging , lmk what you want me to work on next.

abhijeet-dhumal:
Thanks @haroon0x for this, just few nits after this Im happy to unhold once these land. 

/hold

pyproject.toml
abhijeet-dhumal, line 1: Default pytest / `make test-python` will collect tests/benchmarks/ and run pedantic around 100 round benches in every CI job. It seems bit overkill wdyt? 
If possible can we mark/exclude them and add 'make benchmark' (or equivalent) so unit CI stays fast.

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import Any
+
+import pytest
+
+from tests.benchmarks.report import generate_report
+
+DEFAULT_ITERATIONS = 100
+DEFAULT_WARMUP = 5
+OUTPUT_DIR = Path(__file__).parent.parent.parent / "benchmark-results"
+LATENCY_RESULTS: list[dict[str, object]] = []
```
abhijeet-dhumal, line 16: Remaining unresolved: Still needs a reset in pytest_configure so results don’t accumulate across in-process re-runs: LATENCY_RESULTS.clear() (or reassign) at session start..

tests/benchmarks/conftest.py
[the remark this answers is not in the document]
abhijeet-dhumal: Hmm I see.. IIUC benchmark.stats["data"].. that’s valid on pytest-benchmark 5.2

abhijeet-dhumal:
@haroon0x can we resolve above nits? We are very close!
I would like to add this PR in v0.1 milestone 🚀

pyproject.toml
haroon0x: yeah that is overkill , i missed it. I willl exclude benchmark run from the tests.

haroon0x:
@abhijeet-dhumal Lmk if there are any other nits as well

haroon0x:
@abhijeet-dhumal are there any other changes ?

abhijeet-dhumal:
/hold cancel
/lgtm
/approve 
Thanks @haroon0x, great work ! 🚀

PR #27, opened 2026-05-26
fix: detect existing top-level train() to prevent silent training skip

### LINKED ISSUE #19: run_custom_training silently skips training when script defines def train() without calling it

DANIELFRANCIS777:
## Type of Change

- [ ] feat: New feature
- [x] fix: Bug fix
- [ ] revert: Revert a change

## Checklist

## Related Issues
Closes #19

## Problem
`_make_train_func()` always wraps the submitted script in a generated 
`def train():` wrapper. When a user's script already defines a top-level 
`train()`, it becomes nested and is never called — silently skipping training.

## Fix
- Use `ast.parse` to detect top-level `def train()` (sync and async)
- If found and `func_args` match the signature → use script as-is
- If found but `func_args` are missing from signature → raise clear `ValueError`
- If not found → wrap as before (existing behavior unchanged)
- If found but func_args are missing from signature → raise clear 
  ValueError instead of silently failing

## Tests Added
- Plain script gets wrapped correctly
- Async train() detected and executable
- Matching func_args work correctly
- Missing params raise ValueError

Fixes #19


kubeflow_mcp/trainer/api/test_training_utils.py
immkg, line 1: `sdk_contracts_test.py` already exists. It would be better to extend the existing test cases instead of adding a new file.

kubeflow_mcp/trainer/api/training.py
```diff
@@ -103,16 +105,43 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
     When *func_args* is provided, the generated function signature includes matching
     keyword parameters so the trainer can pass them at runtime.
     """
+    # Normalize indentation to handle triple-quoted strings
+    script = textwrap.dedent(script)
+
+    tree = ast.parse(script)
```
immkg: This can raise a `SyntaxError` and currently does not seem to be handled

kubeflow_mcp/trainer/api/training.py
```diff
@@ -103,16 +105,43 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
+            [arg.arg for arg in train_node.args.posonlyargs]
+            + [arg.arg for arg in train_node.args.args]
+            + [arg.arg for arg in train_node.args.kwonlyargs]
+        )
+        # If **kwargs present, skip validation — accepts anything
+        if not train_node.args.kwarg and func_args:
+            missing = [p for p in func_args if p not in param_names]
+            if missing:
+                raise ValueError(f"train() missing required params: {missing}")
+        wrapped = script
```
immkg: The new logic has been added inline in the existing flow. It may be cleaner to encapsulate it into helper functions and call them where needed.

immkg (commented):
Suggested some cleanup around existing test coverage, helper function extraction, and error handling

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: Agreed , wrapping in a  and re-raising as  with a clear message would be the right fix here. Users shouldn't see a raw  bubbling up from .

kubeflow_mcp/trainer/api/test_training_utils.py
abhijeet-dhumal: Agreed. The new cases (async train, kwargs, keyword-only args, missing params) should extend the existing sdk_contracts_test.py rather than live in a separate file. Having two test files for the same function makes it harder to see overall coverage at a glance.

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: +1 on this. Something like _find_top_level_train(tree) and _validate_func_args(train_node, func_args) would make _make_train_func easier to follow and the helpers independently testable.

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: @DANIELFRANCIS777 can you address this ?

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: @DANIELFRANCIS777  can you address this too

kubeflow_mcp/trainer/api/training.py
```diff
@@ -103,16 +105,43 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
```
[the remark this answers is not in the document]
abhijeet-dhumal: @DANIELFRANCIS777 this is worth documenting — when train() already exists, wrapped = script means exec() runs all module-level statements immediately at _make_train_func() call time rather than at train() call time. This is actually more correct for typical ML scripts, but it is a behavioral change from before. A short comment in the code clarifying the intent would prevent confusion.

kubeflow_mcp/trainer/api/training.py
```diff
@@ -103,16 +105,43 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
+            break
+
+    if not has_train:
+        if func_args:
+            params = ", ".join(f"{k}=None" for k in func_args)
+            wrapped_lines = [f"def {func_name}({params}):\n"]
+        else:
+            wrapped_lines = [f"def {func_name}():\n"]
+        for line in lines:
+            wrapped_lines.append(f"    {line}\n")
```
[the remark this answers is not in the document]
abhijeet-dhumal: @DANIELFRANCIS777 this looks legit — if script is empty or whitespace-only after dedent, wrapped_lines has no body lines and compile() will raise SyntaxError. Worth adding a pass fallback.

kubeflow_mcp/trainer/api/training.py
```diff
@@ -103,16 +105,43 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
+        param_names = (
+            [arg.arg for arg in train_node.args.posonlyargs]
+            + [arg.arg for arg in train_node.args.args]
+            + [arg.arg for arg in train_node.args.kwonlyargs]
+        )
+        # If **kwargs present, skip validation — accepts anything
+        if not train_node.args.kwarg and func_args:
+            missing = [p for p in func_args if p not in param_names]
+            if missing:
+                raise ValueError(f"train() missing required params: {missing}")
```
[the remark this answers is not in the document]
abhijeet-dhumal: @DANIELFRANCIS777 this is legit — the variable is named 'missing' but it contains params from func_args that train() does not accept, not params that train() requires. The message reads backwards and will confuse anyone debugging it.

kubeflow_mcp/trainer/api/training.py
```diff
@@ -103,16 +105,43 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
-        wrapped += f"    {line}\n"
+        # Build complete param list from all arg types
+        param_names = (
+            [arg.arg for arg in train_node.args.posonlyargs]
+            + [arg.arg for arg in train_node.args.args]
+            + [arg.arg for arg in train_node.args.kwonlyargs]
+        )
+        # If **kwargs present, skip validation — accepts anything
+        if not train_node.args.kwarg and func_args:
+            missing = [p for p in func_args if p not in param_names]
```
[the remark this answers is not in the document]
abhijeet-dhumal: @DANIELFRANCIS777 this is legit — posonlyargs are included in param_names so validation passes, but func_args are passed as keywords by the trainer at runtime. Calling train(posonly_param=value) will raise TypeError even though validation said it was fine. Rare in practice but worth fixing.

abhijeet-dhumal:
The issue has been fixed and merged !
/close

PR #14, opened 2026-05-11
chore(docs): add SECURITY.md and expand security model in ARCHITECTURE.md

abhijeet-dhumal:
## Description

Adds vulnerability reporting policy and expands security documentation:
- Add SECURITY.md :  vulnerability reporting, disclosure process (aligned with kubeflow/trainer)
- Move threat model, RBAC, hardening checklist to ARCHITECTURE.md's Security Model section
- Fix RBAC manifests: add create/delete/patch verbs for clustertrainingruntimes

## Checklist

- [x] Tests pass locally (`make test-python`)
- [x] Linting passes (`make verify`)
- [x] Documentation updated (if applicable)
- [x] Commit messages follow conventional format

## Related Issues

Fixes: #12

PR #3, opened 2026-04-09
chore: add CI and community hygiene

abhijeet-dhumal:
Fixes https://github.com/kubeflow/mcp-server/issues/8

Hardens CI and community workflows before feature work continues.
Inspired from repo : https://github.com/kubeflow/sdk

PR #30, opened 2026-06-05
feat(ci): replace pip-audit with OSV-Scanner for CVE detection

### LINKED ISSUE #29: Use osv-scanner for CVE detection
### Description

The current `security-audit` job in `test-python.yaml` uses `pip-audit` to detect CVEs.
This works but has limitations, it only queries the PyPI advisory database, has no automated
fix PRs, and doesn't surface findings in the GitHub Security tab.

Replace it with `osv-scanner.yaml` modeled after [kubeflow/sdk's implementation](https://github.com/kubeflow/sdk/blob/main/.github/workflows/osv-scanner.yaml).

### Use Case

_No response_

### Proposed Solution

1. Add `.github/workflows/osv-scanner.yaml` (port from kubeflow/sdk)
2. Remove the `security-audit` job from `test-python.yaml`
3. Remove `pip-audit` from `[dependency-groups] dev` in `pyproject.toml`
4. Keep `[tool.uv] constraint-dependencies` entries until the scanner confirms they're no longer needed

## References

- kubeflow/sdk osv-scanner: https://github.com/kubeflow/sdk/blob/main/.github/workflows/osv-scanner.yaml
- OSV Scanner: https://google.github.io/osv-scanner/
- Tracking: follows the community hygiene work in #8

### Component

Core

Kartikeya-trivedi:
## Description

We were using pip-audit for the security scan, but it's fairly limited ie it only checks PyPI's advisory database, never opens fix PRs, and nothing shows up in GitHub's Security tab. This swaps it out for OSV-Scanner, ported over from kubeflow/sdk.

What changes:
- A nightly scan of `uv.lock` (also runnable on demand from the Actions tab)
- Findings upload as SARIF, so they land in the Security tab
- If a vuln is fixable, it opens a PR bumping the package, a normal upgrade where that works, or a `[tool.uv]` override when a plain upgrade isn't enough

The auto-fix part pulls in three small helper scripts (`extract_version`, `compare_versions`, `update_overrides`). pip-audit is dropped from the dev dependencies and the lockfile is regenerated.

Heads-up, since the repo wasn't quite in the state the issue's task list assumed:
- No `security-audit` job existed in `test-python.yaml` to remove, pip-audit was only a dev dependency, never run in CI.
- pip-audit lived under `[project.optional-dependencies].dev`, not `[dependency-groups].dev`; removed from there.
- No `[tool.uv] constraint-dependencies` section exists yet, so nothing to keep. The scanner only writes `override-dependencies` when it needs to.

For a maintainer with admin access: the auto-fix PR step needs an `area/security` label to exist and "Allow GitHub Actions to create and approve pull requests" enabled (Settings -> Actions -> General). Scanning and the Security-tab upload work without either.

Workflow link = https://github.com/Kartikeya-trivedi/mcp-server/actions/runs/27156215089

## Type of Change

- [x] feat: New feature
- [ ] fix: Bug fix
- [ ] revert: Revert a change

## Checklist

## Related Issues

Fixes #29

### DIFF
--- a/.github/scripts/update_overrides.py
+++ b/.github/scripts/update_overrides.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+"""
+Update override-dependencies in pyproject.toml.
+
+This script manages the [tool.uv] override-dependencies section by:
+1. Reading existing overrides
+2. Adding or updating a package override
+3. Rewriting the entire section in consistent multi-line format
+
+Usage:
+    python update_overrides.py <package> <target> <date> <advisory_url>
+
+Example:
+    python update_overrides.py requests "requests==2.31.0" "2025-01-15" "https://..."
+"""
+
+import re
+import sys
+from pathlib import Path
+
+
+def parse_existing_overrides(content):
+    """Parse existing [tool.uv] override-dependencies into {pkg_name: (spec, comment)}.
+
+    Handles both single-line (`override-dependencies = ["pkg==1.0"]`) and
+    multi-line array formats. Returns an empty dict when no overrides exist.
+    """
+    overrides = {}
+    if "override-dependencies" not in content:
+        return overrides
+
+    in_override = False
+    last_comment = None
+    for line in content.split("\n"):
+        if "override-dependencies" in line and "=" in line:
+            in_override = True
+            # Check if it's a single-line array: override-dependencies = ["pkg1", "pkg2"]
+            single_line_match = re.search(r"override-dependencies\s*=\s*\[(.*?)\]", line)
+            if single_line_match:
+                # Parse all packages from single line
+                for pkg_match in re.finditer(r'"([^"]+)"', single_line_match.group(1)):
+                    spec = pkg_match.group(1)
+                    pkg_name = spec.split("==")[0] if "==" in spec else spec.split("[")[0]
+                    overrides[pkg_name] = (spec, None)
+                break  # Single-line format, done parsing
+            continue
+        if in_override:
+            if line.strip() == "]":
+                break
+            # Check for comment lines
+            if line.strip().startswith("#"):
+                last_comment = line.strip()
+                continue
+            # Extract package spec
+            match = re.search(r'"([^"]+)"', line)
+            if match:
+                spec = match.group(1)
+                pkg_name = spec.split("==")[0] if "==" in spec else spec.split("[")[0]
+                overrides[pkg_name] = (spec, last_comment)
+                last_comment = None
+    return overrides
+
+
+def main():
+    if len(sys.argv) != 5:
+        print(
+            "Usage: update_overrides.py <package> <target> <date> <advisory_url>",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    package = sys.argv[1]
+    target = sys.argv[2]
+    current_date = sys.argv[3]
+    advisory_url = sys.argv[4]
+
+    pyproject_path = Path("pyproject.toml")
+    if not pyproject_path.exists():
+        print(f"Error: {pyproject_path} not found", file=sys.stderr)
+        sys.exit(1)
+
+    # Read current pyproject.toml
+    content = pyproject_path.read_text()
+
+    comment = f"# {target} - Added {current_date} for security fix - {advisory_url}"
+
+    # Extract existing override-dependencies
+    has_tool_uv = "[tool.uv]" in content
+    has_overrides = "override-dependencies" in content
+    overrides = parse_existing_overrides(content)
+
+    # Add or update the current package
+    overrides[package] = (target, comment)
+
+    # Rebuild pyproject.toml
+    if not has_tool_uv:
+        # Add a bare [tool.uv] section at the end; the header comment and
+        # override-dependencies array are inserted by the logic below. (Adding
+        # the header here too would duplicate the whole block -> invalid TOML.)
+        content += "\n[tool.uv]\n"
+        has_tool_uv = True
+
+    if has_overrides:
+        # Remove old override-dependencies section (handles both single-line and multi-line)
+        # Single-line: override-dependencies = ["pkg==1.0"]
+        # Multi-line: override-dependencies = [\n    "pkg",\n]
+        content = re.sub(
+            r"^override-dependencies\s*=\s*\[.*?\]",
+            "",
+            content,
+            flags=re.MULTILINE | re.DOTALL,
+        )
+        # Also remove any orphaned comments before override-dependencies
+        content = re.sub(
+            r"^# Security overrides.*?\n(?:[ \t]*\n)?",
+            "",
+            content,
+            flags=re.MULTILINE,
+        )
+
+        # Re-add the header after [tool.uv] so insertion logic stays consistent
+        content = re.sub(
+            r"(\[tool\.uv\]\r?\n)",
+            r"\1# Security overrides - Review periodically and remove "
+            r"if parent constraints allow natural upgrade\n",
+            content,
+            count=1,
+        )
+
+    # Find [tool.uv] and insert override-dependencies
+    if not has_overrides and has_tool_uv:
+        # Insert after [tool.uv]
+        content = re.sub(
+            r"(\[tool\.uv\]\r?\n)",
+            r"\1# Security overrides - Review periodically and remove if parent constraints allow natural upgrade\n",
+            content,
+        )
+
+    # Build override-dependencies array
+    override_lines = ["override-dependencies = ["]
+    for _pkg_name, (spec, pkg_comment) in sorted(overrides.items()):
+        if pkg_comment:
+            override_lines.append(f"    {pkg_comment}")
+        override_lines.append(f'    "{spec}",')
+    override_lines.append("]")
+    override_block = "\n".join(override_lines)
+
+    # Insert after [tool.uv] or the header comment
+    if "# Security overrides" in content:
+        content = re.sub(r"(# Security overrides.*?\r?\n)", r"\1" + override_block + "\n", content)
+    else:
+        content = re.sub(r"(\[tool\.uv\]\r?\n)", r"\1" + override_block + "\n", content, count=1)
+
+    # Write back
+    pyproject_path.write_text(content)
+    print(f"Updated override-dependencies with {target}")
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/scripts/update_overrides.py
+++ b/.github/scripts/update_overrides.py
@@ -104,15 +104,17 @@ def main():
         # Remove old override-dependencies section (handles both single-line and multi-line)
         # Single-line: override-dependencies = ["pkg==1.0"]
         # Multi-line: override-dependencies = [\n    "pkg",\n]
+        # Leading [ \t]* so an indented key (valid TOML) is still matched and
+        # removed; otherwise the old block survives and we emit a duplicate key.
         content = re.sub(
-            r"^override-dependencies\s*=\s*\[.*?\]",
+            r"^[ \t]*override-dependencies\s*=\s*\[.*?\]",
             "",
             content,
             flags=re.MULTILINE | re.DOTALL,
         )
         # Also remove any orphaned comments before override-dependencies
         content = re.sub(
-            r"^# Security overrides.*?\n(?:[ \t]*\n)?",
+            r"^[ \t]*# Security overrides.*?\n(?:[ \t]*\n)?",
             "",
             content,
             flags=re.MULTILINE,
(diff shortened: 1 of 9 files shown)


.github/scripts/update_overrides.py
```diff
@@ -0,0 +1,160 @@
...
+    if has_overrides:
+        # Remove old override-dependencies section (handles both single-line and multi-line)
+        # Single-line: override-dependencies = ["pkg==1.0"]
+        # Multi-line: override-dependencies = [\n    "pkg",\n]
+        content = re.sub(
+            r"^override-dependencies\s*=\s*\[.*?\]",
+            "",
+            content,
+            flags=re.MULTILINE | re.DOTALL,
+        )
```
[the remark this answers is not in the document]
Kartikeya-trivedi: Good catch I fixed it in b817b11 , now handles indented keys

Krishna-kg732 (changes requested):
Hey @Kartikeya-trivedi  , thanks for picking up this work ! 

overall LGTM  , pretty well aligned with sdk ,
can you also please include a unit test for the scripts as well , like we have in sdk :
https://github.com/kubeflow/sdk/blob/main/.github/scripts/test_scripts.py

edit : Please provide a link to the workflow test run on your fork if possible

cc : @abhijeet-dhumal

Kartikeya-trivedi:
Hi @Krishna-kg732 thanks for the review, I am adding the tests, btw the workflow link is https://github.com/Kartikeya-trivedi/mcp-server/actions/runs/27156215089 will also update it in pr aswell

Krishna-kg732 (commented):
Hey @Kartikeya-trivedi  , **OSV-Scanner version pinning:** The hardcoded `OSV_VERSION="2.3.8"` and checksum require manual monthly updates , I think we should automate this process wdyt?

overall looks good

Kartikeya-trivedi:
HI @Krishna-kg732  yeah agreed, the manual bump is a known annoyance, it's inherited from the sdk version (Dependabot can't bump curl-installed binaries, and we keep the pin + checksum since the workflow runs with write permissions, so blindly pulling latest isn't great supply-chain-wise).

Good news is osv-scanner releases ship an official osv-scanner_SHA256SUMS asset, so this is automatable without losing the checksum verification: a tiny scheduled job that checks the latest release, pulls the version + matching checksum from the SHA256SUMS file, updates the two lines, and opens a PR for review. Same pattern as the security-fix PRs this workflow already creates.

I'd keep this PR aligned with sdk as-is (v2.3.8 is actually the latest release right now, so nothing's stale) and do the auto-bump as a follow-up issue that way it can be contributed to kubeflow/sdk too instead of the two copies drifting. Happy to pick that up. wdyt?

abhijeet-dhumal:
OSV-Scanner replacement for pip-audit is release-critical. Branch has merge conflicts with main — please rebase. Once rebased I'll review.

/assign @Kartikeya-trivedi

Kartikeya-trivedi:
Hi @abhijeet-dhumal, I have rebased my branch and it is free of merge conflicts now, Feel free to review!

Kartikeya-trivedi:
quicknote: I rebased on main - picked up the [dependency-groups] layout, removed the security-audit job that landed with #3, kept the constraint-dependencies pins per the issue.

abhijeet-dhumal:
@Kartikeya-trivedi can you resolve merge conflicts?

abhijeet-dhumal:
Thanks for the contribution, @Kartikeya-trivedi ! 🙏 
A few comments below. Note: this review was performed with the help of AI tools.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Install OSV-Scanner
+        # Bump cadence: check https://github.com/google/osv-scanner/releases monthly
+        # Dependabot cannot bump curl-installed binaries — update version + checksum manually
+        run: |
+          OSV_VERSION="2.3.8"
```
abhijeet-dhumal: Supply-chain risk on checksum drift: The hardcoded OSV_VERSION + EXPECTED_SHA is the right pattern for a workflow with contents: write, but there's no CI gate that alerts when v2.3.8 goes stale. The contributor acknowledged this and plans a follow-up — acceptable for now but should track a follow-up issue explicitly.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
+          if [ "$ACTUAL_SHA" != "$EXPECTED_SHA" ]; then
+            echo "::error::OSV-Scanner checksum mismatch! Expected ${EXPECTED_SHA}, got ${ACTUAL_SHA}"
+            exit 1
+          fi
+          chmod +x osv-scanner
+          sudo mv osv-scanner /usr/local/bin/
+
+      - name: Run OSV-Scanner (SARIF for Security Tab)
+        # continue-on-error so a SARIF failure doesn't block the JSON scan + auto-fix below
+        continue-on-error: true
```
abhijeet-dhumal: continue-on-error: true on the SARIF scan step is correct (exit 1 = vulns found), but the error-handling block re-exits 1 on unexpected codes, which will be swallowed by continue-on-error. Net effect: a genuine scanner crash (e.g. exit 2) is silently ignored. Consider if: always() on the next step + removing continue-on-error, or at minimum emit a ::warning that persists in the Actions summary.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
+            --format json \
+            --output osv-results.json \
+            -L uv.lock || rc=$?
+          if [ "$rc" -ne 0 ] && [ "$rc" -ne 1 ]; then
+            echo "::error::OSV-Scanner failed with unexpected exit code $rc"
+            exit 1
+          fi
+          # Validate JSON output is parseable before downstream steps consume it
+          if [ -f osv-results.json ] && [ -s osv-results.json ]; then
+            jq empty osv-results.json
```
abhijeet-dhumal, line 75: The jq pipeline extracting fix data is complex but reasonable. One edge case: if .references is empty (some OSV entries have none), .references[0].url is null and the fallback works. Good.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
+
+            # Check resulting version
+            TREE_RC=0
+            TREE_OUTPUT=$(uv tree --package "$pkg" 2>&1) || TREE_RC=$?
+            if [ "$TREE_RC" -ne 0 ]; then
+              echo "Warning: uv tree failed for $pkg (exit code $TREE_RC), skipping"
+              continue
+            fi
+            EXTRACT_RC=0
+            NATURAL_VER=$(echo "$TREE_OUTPUT" | uv run python3 .github/scripts/extract_version.py "$pkg") || EXTRACT_RC=$?
```
abhijeet-dhumal: uv run python3 .github/scripts/extract_version.py — this assumes the dev venv has packaging installed (needed by compare_versions.py). Since this runs after astral-sh/setup-uv@v7 + no explicit uv sync, the scripts rely on packaging being available from the lockfile. Should add uv sync --frozen before the fixer step, or the pipeline could fail in a clean runner.

abhijeet-dhumal:
@Kartikeya-trivedi can you confirm and verify workflow adds uv sync before running helper scripts (or confirm uv run auto-syncs in GHA context) ?
and do we need OSV-Scanner auto-bump job  too ?

Kartikeya-trivedi:
Hello @abhijeet-dhumal, thanks for the review, for your questions above yes uv run auto-syncs acc to docs: "uv will ensure that the project environment is up-to-date before running the given command" and the fork run already exercised compare_versions.py (imports packaging) across ~8 packages with no import errors. That said, I'm still applying --with packaging per your other comment, because there's a subtle failure mode: if packaging weren't resolved, the script exits 1 (ModuleNotFoundError), which the fixer would misread as "upgrade sufficient" and silently skip a needed override. --with closes that.

auto-bump job: I'd keep it out of this PR and track it as a follow-up (filing now) — v2.3.8 is current so nothing's stale, this PR is the release unblock so I'd rather not grow its scope, and it's self-contained (same nightly-PR mechanism). I'll mirror it to kubeflow/sdk too so both stay in sync. Can pick it up right after this merges.

.github/workflows/osv-scanner.yaml
Kartikeya-trivedi: Fixed in a67c6dc:- dropped continue-on-error so an unexpected exit code now fails the job instead of being swallowed, and added if: always() to the JSON scan so a SARIF hiccup still doesn't block the independent auto-fix. Exit 0/1 stay green as before.

.github/workflows/osv-scanner.yaml
Kartikeya-trivedi: Fixed in a67c6dc :- compare_versions.py now runs via uv run --with packaging, so it no longer relies on packaging being a transitive lock dep. Also verified uv run auto-syncs (docs + the fork run executed it across ~8 packages with no import error), but the --with closes a subtle hole: a missing packaging exits 1 = ModuleNotFoundError, which the fixer would misread as "sufficient" and silently skip a needed override.

.github/workflows/osv-scanner.yaml
Kartikeya-trivedi: Agreed, tracking as a follow-up (will file an issue for a scheduled auto-bump job that reads the latest release + its osv-scanner_SHA256SUMS, so the checksum verification is preserved). Keeping the manual pin here as the supply-chain guard. Will mirror it to kubeflow/sdk.

abhijeet-dhumal:
/ok-to-test
@Kartikeya-trivedi can you rebase this PR on latest main and resolve merge conflicts?

abhijeet-dhumal:
@Kartikeya-trivedi  Update SECURITY.md ,.. it still says pip-audit runs in CI; should point at the OSV-Scanner nightly workflow after this lands.

Nits (can be same push or quick follow-up):
- Auto-fix PR body: table header + bullet fix_details still won’t render as a table  : emit `| pkg | advisory |` rows or use a list only.
- After merge we’ll need the area/security label + Actions “create PRs” setting for the auto-fix path.

Kartikeya-trivedi:
hi @abhijeet-dhumal , Fixed the table in 2970f30 — the rows now come out as | pkg | advisory | so they actually render inside the table instead of dangling under the header as bullets.

On SECURITY.md — small catch: it's not on main yet, it's coming in via #14, and the "pip-audit runs in CI" line is in that PR's diff. If I touch it here we'd just conflict with #14. if this merges first I'll do a quick 2-line follow-up. 

And yep, the label + "allow Actions to create PRs" setting are called out in the PR description as the post-merge admin checklist, only the auto-fix step needs them, scan + SARIF work either way.

abhijeet-dhumal:
@Kartikeya-trivedi  Almost ready. Direction matches sdk, prior crash/packaging/table issues look fixed. Two blockers before /lgtm:

Two things before `/lgtm`:
1. Wire .github/scripts/test_scripts.py into the Makefile the way [sdk](https://github.com/kubeflow/sdk/blob/35f274cf11bbcf6356a129c624e044eba6e2eb93/Makefile#L150) does (make test-scripts), so CI actually runs them .. right now they only exist on disk.
2. OSV pin is still 2.3.8; latest is v2.4.0. Please bump version + checksum in this PR

pyproject.toml
abhijeet-dhumal, line 1: this file has a pre-existing comment still says “detected by pip-audit”. 
Worth a one-line reword now that we’re on OSV.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
```
abhijeet-dhumal: Nit: `printf '%s\n' "$TREE_OUTPUT" | …` is a bit safer than `echo` for multi-line tree output.

Kartikeya-trivedi: I did as suggested , used printf instead of echo

pyproject.toml
Kartikeya-trivedi: yes did the required change

Kartikeya-trivedi:
Both blockers + nits done:

make test-scripts added (mirrors sdk's target) and wired into the test job right after Verify, so the script tests now gate CI — 8f09847
bumped the pin to v2.4.0 — checksum taken from the release's official SHA256SUMS and cross-verified against the downloaded binary before pinning — 0ad89f6
same commit: printf for the multi-line tree output, and the pyproject comment now says OSV-Scanner instead of pip-audit
should be good for another look 🙏

cc @abhijeet-dhumal

abhijeet-dhumal:
/lgtm
/approve
Thanks @Kartikeya-trivedi for your contribution and addressing recursive reviews promptly 😄 
Great work 🚀

Kartikeya-trivedi:
Thanks @abhijeet-dhumal learned a lot, looking forward to contribute more :)

PR #31, opened 2026-06-06
fix(trainer): execute top-level train calls and strictly validate func_args

### LINKED ISSUE #19: run_custom_training silently skips training when script defines def train() without calling it

reckless-sherixx:
# Description
This PR fixes a silent failure issue where training scripts that define a top-level `train()` function were nested inside the generated wrapper but never executed.

## Type of Change
- [ ] feat: New feature
- [x] fix: Bug fix
- [ ] revert: Revert a change

## Checklist
- [x] Tests pass locally (make test-python)
- [x] Linting passes (make verify)

## Related Issues
Closes #19

## Problem
`_make_train_func()` always wraps the submitted script in a generated `def train():` wrapper. When a user's script already defines a top-level `train()`, it becomes nested and is never called silently skipping training.

## Fix
* Use `ast.parse` to detect top-level `def train()` (sync and async).
* If found and `func_args` match the signature → call it with forwarded arguments.
* If found but `func_args` are missing from the signature or do not match → raise a clear `ValueError` instead of silently failing.
* If not found → wrap as before (existing behavior unchanged).

## Tests Added
* Plain script gets wrapped correctly
* Async `train()` detected and executable
* Matching `func_args` work correctly
* Missing params raise `ValueError`

### DIFF
--- a/kubeflow_mcp/trainer/api/sdk_contracts_test.py
+++ b/kubeflow_mcp/trainer/api/sdk_contracts_test.py
@@ -641,17 +641,42 @@ class TestMCPToSDKConversions:
 
         assert marker == [(0.01, 3)]
 
-    def test_make_train_func_skips_call_for_required_params(self):
-        """train(required_param) with no func_args must NOT auto-append a call."""
-        marker = self._run_wrapped_train(
-            """
-                import builtins
-                def train(required_param):
-                    builtins._kubeflow_mcp_train_marker.append("ran")
-            """
-        )
-        # train() was NOT called because required_param cannot be supplied
-        assert marker == []
+    def test_make_train_func_raises_value_error_for_required_params(self):
+        """train(required_param) with no func_args must raise a ValueError."""
+        with pytest.raises(ValueError, match=r"User-defined train\(\) requires parameters but no func_args were provided"):
+            self._run_wrapped_train(
+                """
+                    import builtins
+                    def train(required_param):
+                        builtins._kubeflow_mcp_train_marker.append("ran")
+                """
+            )
+
+    def test_make_train_func_raises_syntax_error_for_invalid_script(self):
+        """SyntaxError in the user script should be surfaced."""
+        from kubeflow_mcp.trainer.api.training import _make_train_func
+
+        with pytest.raises(SyntaxError, match="Invalid Python script"):
+            _make_train_func("def train():\n    pass\n  invalid syntax")
+
+    def test_make_train_func_raises_value_error_for_mismatched_func_args(self):
+        """train() with no args but func_args provided should raise ValueError."""
+        with pytest.raises(ValueError, match=r"User-defined train\(\) signature does not accept"):
+            self._run_wrapped_train(
+                """
+                    def train():
+                        pass
+                """,
+                func_args={"lr": 0.01}
+            )
+
+    def test_make_train_func_adds_pass_for_empty_script(self):
+        """An empty or whitespace-only script should generate a valid def train(): pass."""
+        from kubeflow_mcp.trainer.api.training import _make_train_func
+
+        train_func = _make_train_func("   \n  \t  \n")
+        assert callable(train_func)
+        train_func()  # Should run without Error
 
     def test_make_train_func_async_train_executes(self):
         """async def train() should be invoked via asyncio.run()."""
--- a/kubeflow_mcp/trainer/api/sdk_contracts_test.py
+++ b/kubeflow_mcp/trainer/api/sdk_contracts_test.py
@@ -641,6 +641,29 @@ class TestMCPToSDKConversions:
 
         assert marker == [(0.01, 3)]
 
+    def test_make_train_func_skips_call_for_required_params(self):
+        """train(required_param) with no func_args must NOT auto-append a call."""
+        marker = self._run_wrapped_train(
+            """
+                import builtins
+                def train(required_param):
+                    builtins._kubeflow_mcp_train_marker.append("ran")
+            """
+        )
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
Thanks @reckless-sherixx ! 
A few small items before /lgtm..

kubeflow_mcp/trainer/api/training.py
```diff
@@ -114,6 +117,10 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
     for line in lines:
         wrapped += f"    {line}\n"
 
+    train_lines = _uncalled_train_call(stripped_script, func_args)
```
abhijeet-dhumal, line 124: ast.parse failures are not being detected here. Consider surfacing a clear SyntaxError("Invalid Python script: ...") from _make_train_func instead of silently skipping the append.

abhijeet-dhumal, line 124: above whitespace-only script builds def train(): with no body and compile might fail. Add pass when lines is empty (same as #27 ).

kubeflow_mcp/trainer/api/training.py
```diff
@@ -124,6 +131,101 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
+        and node.name == "train"
+    )
+    train_def = next(train_defs, None)
+    if train_def is None or _has_train_call(tree.body):
+        return []
+
+    # Guard: if train() has required params and no func_args are provided,
+    # we cannot safely generate a call skip auto-append.
+    if not func_args and _has_required_params(train_def):
```
abhijeet-dhumal, line 161: train(required_param) with no func_args still silently skips training (marker stays empty). Consider raising ValueError at submit time instead of returning a no-op wrapper.

kubeflow_mcp/trainer/api/training.py
```diff
@@ -124,6 +131,101 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
+        return []
+
+    forwarded_args = _forwarded_train_args(train_def, func_args)
```
abhijeet-dhumal, line 166: Looks good. Worth a test that mismatched func_args raises ValueError (e.g. train() + func_args={"lr": 0.01}). wdyt?

kubeflow_mcp/trainer/api/training.py
```diff
@@ -124,6 +131,101 @@ def _make_train_func(script: str, func_args: dict[str, Any] | None = None) -> Ca
...
+        f"train({forwarded_args})" if forwarded_args else "train()"
+    )
+
+    if isinstance(train_def, ast.AsyncFunctionDef):
+        return ["import asyncio", f"asyncio.run({call_str})"]
+    return [call_str]
+
+
+def _has_train_call(nodes: list[ast.stmt]) -> bool:
+    """Return True when module-level code calls ``train()`` directly."""
```
abhijeet-dhumal: Doc: only detects direct train() calls (ast.Name), not aliases like t = train; t().  Would it be good to  document the limitation ?

kubeflow_mcp/trainer/api/sdk_contracts_test.py
```diff
@@ -564,6 +566,104 @@ def train_func():
...
+
+        builtins._kubeflow_mcp_train_marker = []
+        try:
+            train_func = _make_train_func(script, func_args=func_args)
+            train_func(**kwargs)
+            return builtins._kubeflow_mcp_train_marker
+        finally:
+            del builtins._kubeflow_mcp_train_marker
+
+    def test_make_train_func_calls_user_defined_train(self):
```
abhijeet-dhumal, line 585: I would suggest we can add more testcase like : invalid syntax, unexpected func_args -> ValueError, and for empty/whitespace script.

abhijeet-dhumal:
@reckless-sherixx LGTM once the SyntaxError + empty-script handling land.
Please make sure to assign #19 to you if you are planning to work on this .. Thanks!

reckless-sherixx:
@abhijeet-dhumal I have been already assigned this issue and will make the changes

kubeflow_mcp/trainer/api/training.py
reckless-sherixx: yeah i'll document it

kubeflow_mcp/trainer/api/training.py
reckless-sherixx: yeah you are right i'll add a test case about it too

reckless-sherixx:
@abhijeet-dhumal Made the changes you asked for and also added the test cases.

abhijeet-dhumal:
Thanks @reckless-sherixx for the quick turnaround. 
Please close #23 / #27 as superseded on merge.

reckless-sherixx:
@abhijeet-dhumal How  close https://github.com/kubeflow/mcp-server/pull/23 / https://github.com/kubeflow/mcp-server/pull/27 as superseded on merge. I dont have access to do that

PR #45, opened 2026-06-24
fix(trainer): workaround torchtune top-level HF dataset paths

### LINKED ISSUE #32: fine_tune fails: SDK generates dataset.data_dir=/workspace/dataset/. (trailing /.)
### Description

When submitting a fine_tune job with a top-level HuggingFace dataset URI (e.g. hf://tatsu-lab/alpaca), the Kubeflow Trainer SDK constructs dataset.data_dir=/workspace/dataset/. as a torchtune CLI override. torchtune's alpaca_cleaned_dataset passes this as a path component inside HF URIs, producing invalid double-slash paths (hf:///workspace/dataset/./tatsu-lab/alpaca) which cause the training job to crash at dataset load time.

Root cause (SDK-level): kubeflow.trainer.backends.kubernetes.utils.get_args_using_torchtune_config unconditionally calls os.path.join(constants.DATASET_PATH, relative_path) even when relative_path == ".", which Python's os.path.join returns as /workspace/dataset/..

## Fix:
Strip trailing /. and append dataset.source=<local_path> + dataset.data_dir=null so torchtune uses load_dataset(source) against the PVC-resident parquet files directly. The actual fix belongs in the Kubeflow Trainer SDK: if relative_path == ".": args.append(f"dataset.data_dir={DATASET_PATH}") (skip the os.path.join).

### Steps to Reproduce

1. Configure fine_tune with a top-level HF dataset: dataset="hf://tatsu-lab/alpaca"
2. Submit the job
3. Check torchtune logs — dataset load fails with HF URI error

### Expected Behavior

dataset.data_dir=/workspace/dataset (no trailing /.)

### Actual Behavior

dataset.data_dir=/workspace/dataset/. causes torchtune to construct invalid HF URIs

### Version

_No response_

### Python Version

3.11

reckless-sherixx:
Resolves #32

## Description

When submitting a `fine_tune` job with a top-level HuggingFace dataset URI (e.g. `hf://tatsu-lab/alpaca`), the Kubeflow Trainer SDK incorrectly constructs `dataset.data_dir=/workspace/dataset/.`. This trailing `/.` causes `torchtune` to misinterpret the path as a HuggingFace Hub URI (`hf:///workspace/dataset/./...`), crashing the training job at startup.

Since the root cause lives upstream in the Kubeflow Trainer SDK, this PR introduces a targeted workaround inside the `fine_tune` MCP tool. 
- Added a check in `kubeflow_mcp/trainer/api/training.py` to detect top-level HuggingFace URIs.
- Injected a `TrainerArgs` object into the `BuiltinTrainer` options for these URIs to manually override the torchtune CLI arguments with `dataset.source=/workspace/dataset` and `dataset.data_dir=null`.
- This override cleanly forces torchtune to load the PVC-resident parquet files directly, bypassing the bad SDK path.

## Related Issue

Fixes #32

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests


reckless-sherixx:
@abhijeet-dhumal  Kindly review the pr

kubeflow_mcp/trainer/api/training.py
```diff
@@ -865,6 +763,16 @@ def fine_tune(
         if name:
             options.append(Name(name=name))
 
+        # Workaround for SDK bug where top-level HF datasets construct
+        # invalid torchtune dataset.data_dir=/workspace/dataset/.
+        # We override the torchtune CLI args directly via TrainerArgs.
+        if _should_apply_hf_dataset_workaround(dataset):
+            options.append(
+                TrainerArgs(
+                    args=["dataset.source=/workspace/dataset", "dataset.data_dir=null"]
```
abhijeet-dhumal: Should use constant kubeflow.trainer.constants.constants.DATASET_PATH using the SDK  instead of  Hardcoded /workspace/dataset

kubeflow_mcp/trainer/api/training.py
```diff
@@ -865,6 +763,16 @@ def fine_tune(
         if name:
             options.append(Name(name=name))
 
+        # Workaround for SDK bug where top-level HF datasets construct
+        # invalid torchtune dataset.data_dir=/workspace/dataset/.
+        # We override the torchtune CLI args directly via TrainerArgs.
+        if _should_apply_hf_dataset_workaround(dataset):
+            options.append(
```
abhijeet-dhumal, line 874: does TrainerArgs.args get appended after or before the SDK-generated cr.args?
can we cross check TrainerArgs override may conflict with SDK-generated args?

abhijeet-dhumal: The SDK's get_args_using_torchtune_config already emits dataset.data_dir=/workspace/dataset/. via the initializer. The TrainerArgs approach appends additional args — it doesn't replace the SDK-generated ones. Whether OmegaConf last-write-wins semantics apply here depends on how the Trainer admission webhook merges TrainerArgs with the built-in args. If it prepends rather than appends, the SDK's buggy value wins.

kubeflow_mcp/trainer/api/sdk_contracts_test.py
```diff
@@ -969,3 +844,15 @@ def test_fine_tune_sdk_type_params_match(self):
...
             assert p in sig.parameters, f"MCP fine_tune missing LoRA param: {p}"
             assert p in lora_fields, f"SDK LoraConfig missing field: {p}"
+
+    def test_should_apply_hf_dataset_workaround(self):
+        """Workaround helper correctly detects top-level HuggingFace dataset URIs."""
+        from kubeflow_mcp.trainer.api.training import _should_apply_hf_dataset_workaround
+
+        assert _should_apply_hf_dataset_workaround("hf://org/ds") is True
+        assert _should_apply_hf_dataset_workaround("hf://org/ds/") is True
+        assert _should_apply_hf_dataset_workaround("hf://ds") is True
```
abhijeet-dhumal: in this case _should_apply_hf_dataset_workaround accepts hf://ds (single-segment) as True.. 
Single-segment HF URIs are invalid (HF model/dataset IDs are always org/name). This won't cause harm but the test documents incorrect input as valid.

kubeflow_mcp/trainer/api/training.py
abhijeet-dhumal: Missing the dataset.source semantic
dataset.source=/workspace/dataset tells torchtune to use this as the source arg for load_dataset(). But /workspace/dataset is a directory — load_dataset with source=<dir> only works if the directory contains the expected parquet/arrow files. For tatsu-lab/alpaca this works because the initializer downloads the full dataset. For datasets with nested structures (e.g. hf://org/ds/train), the <= 2 check correctly excludes them.

kubeflow_mcp/trainer/api/sdk_contracts_test.py
```diff
@@ -969,3 +844,15 @@ def test_fine_tune_sdk_type_params_match(self):
...
+        assert _should_apply_hf_dataset_workaround("hf://") is False
+        assert _should_apply_hf_dataset_workaround("hf://org/ds/subpath") is False
+        assert _should_apply_hf_dataset_workaround("s3://bucket/dataset") is False
```
abhijeet-dhumal, line 989: No coverage for the confirmed=False preview path

abhijeet-dhumal:
@reckless-sherixx run precommit run to resolve linting issues detected in pr check

reckless-sherixx:
@abhijeet-dhumal I have made changes to uv.lock and docker-publish.yaml cause it was causing ci error .

abhijeet-dhumal:
/lgtm
/approve

Torchtune HF dataset path workaround — full CI green, addresses real training failure mode.

PR #46, opened 2026-06-25
fix(trainer): detect OpenShift pip permission error and document work…

### LINKED ISSUE #41: run_custom_training packages parameter fails on OpenShift (read-only `/.local`)
### Description

When using `run_custom_training(packages=["torch", "transformers"], ...)` on OpenShift, the pre-script pip install step runs `pip install --user` which writes to `/.local`. Under OpenShift's restricted SCC, the root filesystem is read-only and `/.local` is not backed by a writable volume — the job fails immediately with:

`PermissionError: [Errno 13] Permission denied: '/.local'`

The `volumes` parameter **does** auto-inject a writable emptyDir at `/workspace`, but the emptyDir volumes defined by the user (e.g. for `/.local`) are not mounted as volumeMounts on the training container — only `/workspace` is.

### Steps to Reproduce

1. On an OpenShift cluster with restricted SCC
2. Submit:
```python
run_custom_training(
    runtime="torch-distributed",
    script="import torch; print(torch.__version__)",
    packages=["torch"],
    confirmed=True
)
```
3. Check logs → PermissionError: /.local

### Expected Behavior

Packages are installed successfully and the training script runs.

### Actual Behavior

Job fails at pip install with PermissionError.

### Workaround

Do NOT use packages. Install inside the script to the writable /workspace/lib:

```
import subprocess, sys, os
lib_dir = '/workspace/lib'
os.makedirs(lib_dir, exist_ok=True)
subprocess.run([
    sys.executable, '-m', 'pip', 'install',
    '--target', lib_dir, '--quiet',
    'transformers', 'peft', 'trl'
], check=True)
sys.path.insert(0, lib_dir
)
```

### MCP Server Version

_No response_

### Python Version

Python-3.11

### Kubernetes Version

_No response_

### MCP Client

None

### Proposed Fixes

- Documentation: Update platform-fixes.md resource to document this gap and the workaround
- System prompt: Add guidance to INSTRUCTION_SECTIONS["training"] telling agents not to use packages on OpenShift
- Log analysis: Add OpenShift-specific failure pattern to monitoring.py (_FAILURE_PATTERNS) so get_training_logs returns actionable hints
- Long-term: Change the SDK's pip install step to use --target=/workspace/lib instead of --user when running on OpenShift

priyank766:
### Problem 
fixes : https://github.com/kubeflow/mcp-server/issues/41
On OpenShift clusters with a restricted Security Context Constraint (SCC), the root filesystem is read-only. Running `run_custom_training(packages=[...])` executes `pip install --user` (writing to `/.local`), which fails with a `PermissionError: [Errno 13] Permission denied: '/.local'` because user-defined volumes are not mounted during the pre-script installation step.

### Changes
* **Log Parser (`kubeflow_mcp/trainer/api/monitoring.py`)**: Added an OpenShift-specific pattern to `_FAILURE_PATTERNS` to catch permission errors on `/.local` and return the workaround suggestion.
* **Agent Instructions (`kubeflow_mcp/trainer/__init__.py`)**: Added prompt guidance under `INSTRUCTION_SECTIONS["training"]` advising AI agents to avoid the `packages` parameter on OpenShift.
* **Documentation (`kubeflow_mcp/trainer/resources/platform-fixes.md`)**: Documented the issue and the target-directory workaround.
* **Tests (`tests/unit/trainer/test_monitoring.py`)**: Added a new test suite to verify the parser correctly matches and prioritizes the OpenShift error over generic permission errors.

### Workaround
Do not use the `packages` parameter on OpenShift. Instead, run the installation within the training script using:
```python
import subprocess, sys, os
lib_dir = '/workspace/lib'
os.makedirs(lib_dir, exist_ok=True)
subprocess.run([sys.executable, '-m', 'pip', 'install', '--target', lib_dir, '--quiet', 'transformers', 'peft', 'trl'], check=True)
sys.path.insert(0, lib_dir)
```

### Verification
* All 147 unit tests pass.
* Ruff checks and formatting pass with zero violations.

### DIFF
--- a/kubeflow_mcp/trainer/resources/platform-fixes.md
+++ b/kubeflow_mcp/trainer/resources/platform-fixes.md
@@ -96,7 +96,24 @@ For `fine_tune()`, these volumes apply to ALL replicated jobs (node, dataset-ini
 "env": {"NCCL_P2P_DISABLE": "1", "NCCL_SHM_DISABLE": "1"}
 ```
 
-**4. If emptyDirs are not enough** — escalate to cluster admin:
+**4. Pip package installation permission denied in run_custom_training()** — when using `run_custom_training(packages=[...])` on OpenShift under a restricted SCC, the SDK's pre-script pip install step attempts to run `pip install --user` which writes to `/.local`. Since user-defined emptyDir volumes are not mounted on the training container during this step, the job will fail immediately with:
+`PermissionError: [Errno 13] Permission denied: '/.local'`
+
+*Workaround*: Do **NOT** use the `packages` parameter on OpenShift. Instead, write a workaround directly in your training script to install the required packages to `/workspace/lib` (which is a writable emptyDir) and append it to `sys.path`:
+
+```python
+import subprocess, sys, os
+lib_dir = '/workspace/lib'
+os.makedirs(lib_dir, exist_ok=True)
+subprocess.run([
+    sys.executable, '-m', 'pip', 'install',
+    '--target', lib_dir, '--quiet',
+    'transformers', 'peft', 'trl'
+], check=True)
+sys.path.insert(0, lib_dir)
+```
+
+**5. If emptyDirs are not enough** — escalate to cluster admin:
 
 ```bash
 oc adm policy add-scc-to-user anyuid -z <service-account> -n <namespace>
--- a/kubeflow_mcp/trainer/resources/platform-fixes.md
+++ b/kubeflow_mcp/trainer/resources/platform-fixes.md
@@ -18,7 +18,8 @@ Error-based detection (from `get_training_logs()`):
 | Error pattern | Platform | Fix |
 |---------------|----------|-----|
 | "Read-only file system" | OpenShift (restricted SCC) | Add emptyDir volumes below |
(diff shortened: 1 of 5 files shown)


abhijeet-dhumal:
/lgtm

OpenShift pip permission detection is exactly what we need. Consider consolidating with #47 if both touch the same error path.

kubeflow_mcp/trainer/resources/platform-fixes.md
abhijeet-dhumal, line 1: @priyank766  One doc nit: the detection table + "/.local emptyDir fixes most pip issues" still send agents to emptyDir for Permission denied on /.local, but new version says the packages= pre-script never gets those mounts. 
Can you split that case in the table (runtime writes to emptyDir; packages= to /workspace/lib workaround) so agents don't mis-route?
WDYT?

priyank766: Yeah Makes sense. I've split the error cases in both `platform-fixes.md` and `troubleshooting.md` to separate runtime write issues from the packages pre-script limitation. 
@abhijeet-dhumal

abhijeet-dhumal:
/lgtm
/approve
Thanks @priyank766 for your contribution, this is great work 🚀

PR #47, opened 2026-06-25
fix(trainer): fallback to previous pod logs on crash and detect OpenS…

### LINKED ISSUE #42: get_training_logs returns empty for crash-looped / terminated pods
### Description

When a training job's pods crash-loop or terminate (common on OpenShift with permission errors), `get_training_logs()` returns empty logs via the Trainer SDK. The SDK's log fetching doesn't fall back to `previous: true` container logs, so there's no way to see *why* the pod crashed.
Additionally, when logs ARE returned, common OpenShift failure patterns (Permission denied on /.local, HF cache write failures) aren't detected — the raw logs are returned with no actionable guidance.

### Steps to Reproduce

1. Submit a training job that crashes (e.g. missing HF_HOME on OpenShift)
2. Call `get_training_logs(name="<job-name>")`
3. Receive empty or unhelpful logs

### Expected Behavior

1. Logs from crashed pods are retrieved (via `previous=True` fallback to CoreV1Api)
2. Known failure patterns are detected and actionable hints are returned

### 1. CoreV1Api fallback for pod logs
When the Trainer SDK returns empty logs, fall back to `CoreV1Api.read_namespaced_pod_log()` with `previous=True` to fetch logs from the previous container instance:
```python
pods = v1.list_namespaced_pod(namespace=ns, label_selector=f"training.kubeflow.org/trainjob-name={name}")
for pod in pods.items:
    raw = v1.read_namespaced_pod_log(name=pod.metadata.name, namespace=ns, previous=True)
```

### MCP Server Version

_No response_

### Python Version

3.11

### Kubernetes Version

_No response_

### MCP Client

Cursor

### References

_No response_

reckless-sherixx:
Resolves #42

## Description

When a training job's pods crash-loop or terminate (common on OpenShift environments due to random UID permission gotchas or HuggingFace cache errors), `get_training_logs()` previously returned empty logs via the Trainer SDK because active container logs are flushed upon crash/restart.

This PR implements:
1. **`previous=True` Fallback**: In `get_training_logs()`, if active container logs are empty, the server queries `CoreV1Api.read_namespaced_pod_log(..., previous=True)` to recover the crash output from terminated container instances.
2. **OpenShift & HF Failure Hints**: Adds error signature pattern matching for random UID home directory write errors (`Permission denied` on `/.local`, `/home`) and HuggingFace cache directory failures, returning actionable guidance (`set HF_HOME=/workspace`).

## Related Issue

Fixes #42

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

Added `test_extract_failure_hint_openshift_and_hf` and `test_get_training_logs_fallback_to_previous_logs` in `sdk_contracts_test.py` verifying pattern extraction and mock pod log fallback.

### DIFF
--- a/kubeflow_mcp/trainer/api/monitoring.py
+++ b/kubeflow_mcp/trainer/api/monitoring.py
@@ -40,16 +40,6 @@ MIN_POLLING_INTERVAL = 1
 
 _FAILURE_PATTERNS: list[tuple[re.Pattern[str], str, str]] = [
     (
-        re.compile(r"Permission denied.*(/\.local|/home|/\.cache)", re.IGNORECASE),
-        "OPENSHIFT_PERMISSION_ERROR",
-        "OpenShift random UID cannot write to home directory; set env var HF_HOME=/workspace or mount a writable volume.",
-    ),
-    (
-        re.compile(r"Permission denied.*huggingface|HF_HOME", re.IGNORECASE),
-        "HF_CACHE_WRITE_ERROR",
-        "Set env var HF_HOME=/workspace to store HuggingFace cache on a writable volume mount.",
-    ),
-    (
         re.compile(r"CUDA out of memory", re.IGNORECASE),
         "OOM",
         "Reduce batch_size, enable quantization (int8/int4), or request a larger GPU.",
@@ -78,6 +68,16 @@ _FAILURE_PATTERNS: list[tuple[re.Pattern[str], str, str]] = [
         re.compile(r"FileNotFoundError", re.IGNORECASE),
         "FILE_NOT_FOUND",
         "Check dataset/model paths and volume mounts.",
+    ),
+    # HF cache pattern MUST come before the generic PermissionError catch-all
+    # so that /.cache/huggingface failures are classified correctly.
+    (
+        re.compile(
+            r"Permission denied.*(?:huggingface|HF_HOME)|(?:huggingface|HF_HOME).*Permission denied",
+            re.IGNORECASE,
+        ),
+        "HF_CACHE_WRITE_ERROR",
+        "Set env var HF_HOME=/workspace to store HuggingFace cache on a writable volume mount.",
     ),
     (
         re.compile(
--- a/kubeflow_mcp/trainer/api/monitoring.py
+++ b/kubeflow_mcp/trainer/api/monitoring.py
@@ -113,6 +113,20 @@ def _extract_failure_hint(logs: str) -> dict[str, str] | None:
     return None
 
 
+def _is_pod_for_step(pod: Any, step: str) -> bool:
+    """Return whether a JobSet pod corresponds to the requested TrainJob step."""
+    labels = getattr(pod.metadata, "labels", None) or {}
+    replicated_job = labels.get("jobset.sigs.k8s.io/replicatedjob-name")
+    job_index = labels.get("jobset.sigs.k8s.io/job-index")
+    if replicated_job == step:
+        return True
+    return (
+        replicated_job is not None
+        and job_index is not None
+        and f"{replicated_job}-{job_index}" == step
+    )
+
+
 def get_training_logs(
     name: str,
     step: str = "node-0",
@@ -167,6 +181,8 @@ def get_training_logs(
                     label_selector=f"training.kubeflow.org/trainjob-name={name}",
                 )
                 for pod in pods.items:
+                    if not _is_pod_for_step(pod, step):
+                        continue
                     try:
                         raw = v1.read_namespaced_pod_log(
                             name=pod.metadata.name,
--- a/kubeflow_mcp/trainer/api/monitoring.py
+++ b/kubeflow_mcp/trainer/api/monitoring.py
@@ -14,13 +14,24 @@
 
 """Monitoring tools for training job logs and events."""
 
+import logging
 import re
 from typing import Any
 
 from kubeflow_mcp.common.constants import ErrorCode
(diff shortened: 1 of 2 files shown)


reckless-sherixx:
@abhijeet-dhumal Kindly review the pr i have made the changes suggested by copilot too

kubeflow_mcp/trainer/api/sdk_contracts_test.py
```diff
@@ -969,3 +844,45 @@ def test_fine_tune_sdk_type_params_match(self):
...
+        """Verify _extract_failure_hint detects OpenShift permission and HF cache errors."""
+        from kubeflow_mcp.trainer.api.monitoring import _extract_failure_hint
+
+        hint1 = _extract_failure_hint("Permission denied: '/.local/lib'")
+        assert hint1 is not None and hint1["category"] == "OPENSHIFT_PERMISSION_ERROR"
+
+        hint2 = _extract_failure_hint("Permission denied when writing huggingface cache")
+        assert hint2 is not None and hint2["category"] == "HF_CACHE_WRITE_ERROR"
+
+    def test_get_training_logs_fallback_to_previous_logs(self):
```
Krishna-kg732, line 1102: Assert fallback trigger condition explicitly

In test_get_training_logs_fallback_to_previous_logs, ensure test verifies fallback happens only when get_job_logs returns empty.

Krishna-kg732 (changes requested):
Hey @reckless-sherixx , thanks for working on this , overall looks good to me , just a tiny nit 

cc : @abhijeet-dhumal

abhijeet-dhumal:
/lgtm

Clean fix for crash-looped pod log retrieval. Overlaps partially with #46 — if both land, verify no duplicate error-detection paths.

abhijeet-dhumal:
@reckless-sherixx can you resolve merge conflicts ?

reckless-sherixx:
> @reckless-sherixx can you resolve merge conflicts ?

Yeah sure I'll do that

kubeflow_mcp/trainer/api/monitoring.py
```diff
@@ -14,20 +14,41 @@
...
 MAX_EVENT_LIMIT = 500
 MAX_WAIT_TIMEOUT = 3600
 MIN_POLLING_INTERVAL = 1
 
 _FAILURE_PATTERNS: list[tuple[re.Pattern[str], str, str]] = [
+    (
+        re.compile(r"Permission denied.*(/\.local|/home|/\.cache)", re.IGNORECASE),
+        "OPENSHIFT_PERMISSION_ERROR",
+        "OpenShift random UID cannot write to home directory; set env var HF_HOME=/workspace or mount a writable volume.",
+    ),
```
Krishna-kg732: Hey @reckless-sherixx  , Can you please check with @priyank766 's implementation of the same in #46 

https://github.com/kubeflow/mcp-server/pull/46/changes#diff-e505ca5930a15e63cca967e22f9a088d25156c30765aa80d3222e596695c483dR61

kubeflow_mcp/trainer/api/monitoring.py
```diff
@@ -14,20 +14,41 @@
...
+    (
+        re.compile(r"Permission denied.*huggingface|HF_HOME", re.IGNORECASE),
```
abhijeet-dhumal: @reckless-sherixx  HF regex still seems broken
Can we put the HF pattern before the general OpenShift one so /.cache HF failures don’t become OPENSHIFT_PERMISSION_ERROR.

abhijeet-dhumal: something like this ? WDYT?
```
r"Permission denied.*(?:huggingface|HF_HOME)|(?:huggingface|HF_HOME).*Permission denied"
```

abhijeet-dhumal: And test coverage should track this correctly

abhijeet-dhumal:
@reckless-sherixx #46 just merged .. it lands OPENSHIFT_PIP_ERROR for Permission denied on /.local 
Please rebase onto main, and take actions accordingly for clean fix rebase..
cc @Krishna-kg732 (your earlier #46 note)

reckless-sherixx:
Ok I'll make the changes

Krishna-kg732:
/lgtm 

thanks for working on this 

cc: @abhijeet-dhumal , @jaiakash

abhijeet-dhumal:
@reckless-sherixx can you clean rebase on latest main to resolve merge conflicts?

kubeflow_mcp/trainer/api/monitoring.py
```diff
@@ -134,6 +155,42 @@ def get_training_logs(
 
         client = get_trainer_client_for_namespace(namespace)
         log_lines = list(client.get_job_logs(name=name, step=step, follow=False))
+        if not log_lines:
```
abhijeet-dhumal, line 172: nit: IIUC this walks every TrainJob pod and concatenates previous logs, but get_job_logs() above is scoped to 'step' (default node-0). On multi-node jobs we can mix workers.

Maybe worth filtering to the step pod (name/label match on `step`) before read_namespaced_pod_log

but it's fine as follow-up if you want to keep this PR small.

reckless-sherixx:
@abhijeet-dhumal addressed the conflict and also addressed the nit

abhijeet-dhumal:
/lgtm
/approve
Thanks @reckless-sherixx for your contribution and addressing all reviews promptly 😄
Great work 🚀

reckless-sherixx:
@abhijeet-dhumal 😁

PR #43, opened 2026-06-24
fix: suggest valid HuggingFace model IDs when the format check fails

AhmedDlshad007:
# Description
`pre_flight` (and `estimate_resources`, which share `_get_model_info_from_hf`) returned a bare `Invalid HuggingFace model ID format` error for a malformed model reference, with no guidance. Users and AI agents then retry with similarly wrong IDs, wasting round-trips. This attaches best-effort suggestions of real Hub model IDs, following the approach proposed in the issue.

## Type of Change
- [ ] feat: New feature
- [x] fix: Bug fix
- [ ] revert: Revert a change
- [ ] chore: Maintenance / tooling

## Checklist
- [x] Tests pass locally (make test-python)
- [x] Linting passes for the changed files (ruff check + ruff format, under both ruff 0.14.14 and 0.15.12); see the CI note below
- [ ] Documentation updated (if applicable)
- [x] Commit messages follow conventional format

## Related Issues
Fixes #40

## Problem
With an invalid model id (e.g. an Ollama-style tag `qwen3:8b`, or an `hf://` prefixed value), `_get_model_info_from_hf` fails the `_HF_MODEL_ID_RE` check and returns only:
```json
{"error": "Invalid HuggingFace model ID format: 'qwen3:8b'"}
```

## Fix
Add a small `_suggest_hf_model_ids(model)` helper that:
* Normalizes the input into a search term (drops an `hf://` prefix and any Ollama-style `:tag` suffix).
* Queries `huggingface_hub.list_models(search=..., limit=3, full=False)` for close matches.
* Attaches them as a `suggestions` list on the error response.

It returns no suggestions (and adds no `suggestions` key) when the Hub lookup errors or finds nothing, so the validation path stays robust offline and never depends on network access. `huggingface_hub` is already a dependency.

Result now:
```json
{"error": "Invalid HuggingFace model ID format: 'qwen3:8b'", "suggestions": ["Qwen/Qwen3-8B", "Qwen/Qwen2.5-7B-Instruct"]}
```

## Tests Added
New `tests/unit/trainer/test_planning.py` (mocks `list_models`, no network):
* Ollama `:tag` and `hf://` prefix are normalized before searching the Hub.
* Suggestions attached on invalid format; the key is omitted when none are found or the Hub errors.
* Blank-after-normalization input makes no Hub call.

Full suite passes locally: 151 passed (144 existing + 7 new).

## Note on CI
[...]

Issue #40, opened 2026-06-24
pre_flight should suggest valid HuggingFace model IDs on format validation failure


abhijeet-dhumal:
### Description

When `pre_flight()` is called with an invalid HuggingFace model ID (e.g. Ollama-format `qwen3:8b`, or a typo like `meta-lama/Llama-3`), the response is a bare error with zero guidance:
```json
{"error": "Invalid HuggingFace model ID format: 'qwen3:8b'"}
```

Users and AI agents retry with similarly wrong IDs, wasting round-trips.

### Steps to Reproduce

pre_flight(model="qwen3:8b") or
pre_flight(model="meta-lama/Llama-3")

### Expected Behavior

{
  "error": "Invalid HuggingFace model ID format: 'qwen3:8b'",
  "suggestions": ["Qwen/Qwen3-8B", "Qwen/Qwen2.5-7B-Instruct"]
}

### Actual Behavior

{"error": "Invalid HuggingFace model ID format: 'qwen3:8b'"}

### Python Version

3.11

### MCP Client

Cursor

abhijeet-dhumal:
## Proposed Fix
After the regex validation failure in planning.py :

- Normalise the raw input — strip hf:// prefix, Ollama tags (:8b), trailing numeric suffixes
- Query huggingface_hub.list_models(search=<normalised>, limit=3, full=False)
- Attach matching model IDs as a suggestions list on the error response

huggingface_hub is already a transitive dependency so no new deps needed.

```
suggestions = _suggest_hf_model_ids(model)
result = {"error": f"Invalid HuggingFace model ID format: '{model}'"}
if suggestions:
    result["suggestions"] = suggestions
return result
```

AhmedDlshad007:
I'd like to work on this, following your proposed fix.

Plan: after the `_HF_MODEL_ID_RE` check fails in `planning.py`, add a small `_suggest_hf_model_ids(model)` helper that normalizes the raw input (strip the `hf://` prefix and any Ollama-style `:tag` suffix), queries `huggingface_hub.list_models(search=<normalized>, limit=3, full=False)`, and attaches the matching IDs as a `suggestions` list on the error response. It degrades gracefully (no `suggestions` key) when the Hub lookup errors or returns nothing, so the validation path stays robust offline. I'll cover it with unit tests that mock `list_models`.

/assign

sumitraj7766:
[...]

PR #66, opened 2026-07-12
feat: rank model-id suggestions by downloads and suggest on the not-found path

### LINKED ISSUE #65: Follow-ups from #43: suggest on the not-found path, and prefer canonical repos in ranking
Splitting out the two follow-ups from #43 (fix for #40), as suggested by @abhijeet-dhumal.

1. **Suggest on the not-found path too.** Today the "did you mean" suggestions only kick in when a model id fails the format check (like `qwen3:8b`). A valid-looking but nonexistent id such as `meta-lama/Llama-3` goes down the not-found path instead and comes back with just the raw Hub error. Running the same normalization and search there would give typos a suggestion list as well.

2. **Prefer canonical repos in the ranking.** `_suggest_hf_model_ids` uses `list_models(search=...)` with the Hub's default ordering, so community fine-tunes can land above the official model. Passing `sort="downloads"` would put the canonical repos first.

Both are small and non-breaking, and the tests in `test_planning.py` already give a seam to extend. I'll pick this up.

AhmedDlshad007:
# Description
Follow-up to #43 (fix for #40), implementing the two enhancements from #65. Both improve the HuggingFace model-id suggestions that `pre_flight` / `estimate_resources` return for a bad model id.

## Type of Change
- [x] feat: New feature
- [ ] fix: Bug fix
- [ ] revert: Revert a change

## Checklist
- [x] Tests pass locally (184 passing)
- [x] Linting passes (ruff check + ruff format + pre-commit, under both ruff 0.14.14 and 0.15.x)

## Related Issues
Fixes #65

## Changes
1. **Rank suggestions by downloads.** `_suggest_hf_model_ids` now passes `sort="downloads"` to `list_models`, so canonical repos surface ahead of community fine-tunes.
2. **Suggest on the not-found path too.** A valid-format but nonexistent id (a typo like `meta-lama/Llama-3`) previously hit the `except` branch in `_get_model_info_from_hf` and returned only the raw Hub error. It now reuses `_suggest_hf_model_ids`. Since a typo in the org makes the full `org/name` search come back empty, the helper retries on just the model name so the user still gets a "did you mean".

## Verified end to end (live Hub)
```
estimate_resources("qwen3:8b")           # format branch, ranked
  -> ['Qwen/Qwen3-0.6B', 'Qwen/Qwen3-8B', 'Qwen/Qwen3-Embedding-0.6B']   (was community fine-tunes)

estimate_resources("meta-lama/Llama-3")  # not-found path, org typo -> name fallback
  -> ['meta-llama/Llama-3.2-1B-Instruct', 'meta-llama/Llama-3.1-8B-Instruct', 'meta-llama/Llama-3.2-3B-Instruct']
```

## Tests Added
Five new tests in `test_planning.py` (mocking `list_models` / `model_info`): download ranking, the org-typo name fallback, not-found suggestions at both the helper and the `estimate_resources` boundary, and graceful omission when the Hub errors.

### DIFF
--- a/kubeflow_mcp/trainer/api/planning.py
+++ b/kubeflow_mcp/trainer/api/planning.py
@@ -31,22 +31,40 @@ logger = logging.getLogger(__name__)
 _HF_MODEL_ID_RE = re.compile(r"^[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+$")
 
 
+_MIN_NAME_FALLBACK_LEN = 3
+
+
 def _suggest_hf_model_ids(model: str, limit: int = 3) -> list[str]:
     """Suggest real HuggingFace model IDs for a malformed model reference.
 
     Normalizes common non-Hub formats into a search term (drops an ``hf://``
     prefix and any Ollama-style ``:tag`` suffix such as ``qwen3:8b``) and asks
-    the Hub for close matches. Returns an empty list when the lookup errors or
-    finds nothing, so the suggestions stay best-effort and the validation path
-    never depends on network access.
+    the Hub for close matches ranked by downloads, so canonical repos surface
+    ahead of community fine-tunes. If a full ``org/name`` term finds nothing
+    (usually a typo in the org, e.g. ``meta-lama/Llama-3``), it retries on just
+    the model name; that retry is skipped when the name is very short, since a
+    bare one or two character search is too broad to be a useful "did you mean".
+    Returns an empty list when the lookup errors or finds nothing, so the
+    suggestions stay best-effort and the validation path never depends on
+    network access.
     """
-    search = model.strip().removeprefix("hf://").split(":", 1)[0].strip()
-    if not search:
+    normalized = model.strip().removeprefix("hf://").split(":", 1)[0].strip()
+    if not normalized:
         return []
+    terms = [normalized]
+    name = normalized.rsplit("/", 1)[-1]
+    if name and name != normalized and len(name) >= _MIN_NAME_FALLBACK_LEN:
+        terms.append(name)
     try:
         from huggingface_hub import list_models
 
-        return [m.id for m in list_models(search=search, limit=limit, full=False)]
+        for term in terms:
+            matches = [
+                m.id for m in list_models(search=term, sort="downloads", limit=limit, full=False)
+            ]
+            if matches:
+                return matches
+        return []
     except Exception:  # noqa: BLE001 - suggestions are best-effort, never fatal
         return []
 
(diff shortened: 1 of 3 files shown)


kubeflow_mcp/trainer/api/planning.py
```diff
@@ -82,7 +95,14 @@ def _get_model_info_from_hf(model: str) -> dict[str, Any] | None:
         }
 
     except Exception as e:
-        return {"error": str(e)}
+        # A valid-format but nonexistent id (e.g. a typo) lands here rather than
+        # the format branch above; offer the same best-effort suggestions so
+        # callers still get a "did you mean" list.
+        result: dict[str, Any] = {"error": str(e)}
+        suggestions = _suggest_hf_model_ids(model)
```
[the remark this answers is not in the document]
AhmedDlshad007: Narrowed the suggestion lookup to `huggingface_hub` `RepositoryNotFoundError`, so timeouts, rate-limits, and metadata errors now surface unchanged with no extra Hub request. Confirmed that gated repos (Llama/Gemma) return metadata from `model_info` without raising, so they never reach this branch. Added a test asserting a non-not-found error (a timeout) attaches no suggestions and makes no `list_models` call.

kubeflow_mcp/trainer/api/planning.py
abhijeet-dhumal, line 1: Nit: This file has some pre-existing comments still saying suggestions come from the format check. They also attach on RepositoryNotFoundError now .. maybe worth a one-line update. WDYT?

kubeflow_mcp/trainer/api/planning.py
```diff
@@ -36,17 +36,30 @@ def _suggest_hf_model_ids(model: str, limit: int = 3) -> list[str]:
...
+    nothing, so the suggestions stay best-effort and the validation path never
+    depends on network access.
     """
-    search = model.strip().removeprefix("hf://").split(":", 1)[0].strip()
-    if not search:
+    normalized = model.strip().removeprefix("hf://").split(":", 1)[0].strip()
+    if not normalized:
         return []
+    terms = [normalized]
+    name = normalized.rsplit("/", 1)[-1]
```
abhijeet-dhumal, line 55: Nit only: when the org typo path falls back to just the model name (e.g. "Llama-3"), Hub search can be broad. 
limit=3 + sort=downloads can keep it usable.. wdyt?

abhijeet-dhumal: Btw optionally we can skip the name-only retry when len(name) is very short.

abhijeet-dhumal:
@AhmedDlshad007 are you still working on this? 
This PR needs clean rebase..

AhmedDlshad007:
Sorry for the slow reply on this. Rebased onto latest main now (clean, single commit). I also addressed the fallback nits: the name-only retry keeps `sort=downloads` + `limit=3`, and it now skips when the model name is very short. Ready for another look whenever you get a chance.

abhijeet-dhumal:
Thanks @AhmedDlshad007 for this contribution 🚀 
Looks good after the rebase.
/lgtm
/approve

PR #90, opened 2026-07-18
feat(ci): add weekly OSV-Scanner version bump workflow

Kartikeya-trivedi:
## Description

Follow-up from the checksum-drift review thread on #30: the pinned `OSV_VERSION` + `EXPECTED_SHA` is the right supply-chain pattern for a `contents: write` workflow, but bumping it was a manual monthly chore and nothing alerted when the pin went stale.

This adds a weekly workflow (Mondays 03:00 UTC + `workflow_dispatch`) that:

- checks the latest `google/osv-scanner` release, and exits quietly if we're already on it
- extracts the `linux_amd64` checksum from the release's official `osv-scanner_SHA256SUMS`
- cross-verifies that checksum against the actually-downloaded binary before trusting it
- rewrites the `OSV_VERSION` / `EXPECTED_SHA` lines in `osv-scanner.yaml` and opens a PR for human review

The pin + checksum guard stays exactly as it is — this only removes the manual chore. A human still reviews every bump PR before it merges.

Validated the plumbing end-to-end against real releases: the extraction reproduces the v2.3.8 checksum we originally pinned byte-for-byte, and correctly detected the 2.3.8 -> 2.4.0 bump (which #30 then picked up manually). Since #30 merged with 2.4.0 pinned, this workflow's first runs will no-op until the next osv-scanner release.

Note: the bump PRs this creates need the same two admin settings as the auto-fix PRs from #30 (`area/security` label + Actions "allow create PRs").

Planning to mirror this to kubeflow/sdk as well, since it has the same manual-bump chore.

## Type of Change

- [x] feat: New feature
- [ ] fix: Bug fix
- [ ] revert: Revert a change

## Checklist

## Related Issues

Follow-up to #30 (checksum-drift review thread)


.github/workflows/osv-scanner-bump.yaml
```diff
@@ -0,0 +1,90 @@
...
+          WORKFLOW=".github/workflows/osv-scanner.yaml"
+          sed -i "s/OSV_VERSION=\"${{ steps.check.outputs.current }}\"/OSV_VERSION=\"${{ steps.check.outputs.latest }}\"/" "$WORKFLOW"
+          sed -i "s/EXPECTED_SHA=\"[a-f0-9]\{64\}\"/EXPECTED_SHA=\"${{ steps.checksum.outputs.sha }}\"/" "$WORKFLOW"
+          git diff --stat
+
+      - name: Create Pull Request
+        if: steps.check.outputs.update_needed == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
Kartikeya-trivedi: good catch, this was a real design bug: workflows isn't even a grantable scope in the permissions: block, so every actual bump would have died at push. The token line is now secrets.WORKFLOW_BUMP_TOKEN || secrets.GITHUB_TOKEN with a comment documenting what the secret needs (fine-grained PAT or GitHub App token with contents + pull-requests + workflows write). Until that secret exists, weekly runs stay green on the no-op path and an actual bump fails loudly at push instead of silently.

.github/workflows/osv-scanner-bump.yaml
```diff
@@ -0,0 +1,90 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          WORKFLOW=".github/workflows/osv-scanner.yaml"
+          CURRENT=$(sed -n 's/.*OSV_VERSION="\([^"]*\)".*/\1/p' "$WORKFLOW" | head -1)
+          LATEST=$(gh api repos/google/osv-scanner/releases/latest --jq '.tag_name' | sed 's/^v//')
+          echo "current=$CURRENT latest=$LATEST"
+          if [ -z "$CURRENT" ] || [ -z "$LATEST" ]; then
+            echo "::error::Could not determine current ($CURRENT) or latest ($LATEST) version"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
Kartikeya-trivedi: both versions are now validated as strict X.Y.Z before they're written to step outputs, so nothing with shell/sed metacharacters can reach the later interpolations. The rewrite helper independently re-validates its inputs too (tested with a literal 9.9.9"; rm -rf / — rejected).

Kartikeya-trivedi:
@abhijeet-dhumal this is the follow-up I promised on the checksum-drift thread in #30, the weekly auto-bump for the OSV-Scanner pin, so the version + checksum never go stale silently again. Same trust model as before: checksum comes from the release's official SHA256SUMS, gets cross-verified against the downloaded binary, and a human still reviews every bump PR before merge.

Copilot's review already caught some good stuff (token scope, input validation, downgrade guard), all addressed in 07f4d0d, with the rewrite logic extracted into a tested helper per the repo convention we set up in #30.

Would appreciate a review when you get a chance 🙏

.github/workflows/osv-scanner-bump.yaml
Krishna-kg732, line 1: rename the file to be consistent with the sdk repo's workflow for the same , ie; osv-scanner.yaml

.github/workflows/osv-scanner-bump.yaml
```diff
@@ -0,0 +1,112 @@
+name: OSV-Scanner Version Bump
```
Krishna-kg732, line 1: check the sdk's implementation of the same at : https://github.com/kubeflow/sdk/blob/main/.github/workflows/osv-scanner.yaml

Krishna-kg732 (changes requested):
Hey @Kartikeya-trivedi thanks for picking this up, can you please take a look at how the sdk handles the version bumps using osv scanner

Krishna-kg732 (commented):
Hey @abhijeet-dhumal , do we want this workflow similar to how we have it in SDK? wdyt

.github/workflows/osv-scanner-bump.yaml
abhijeet-dhumal: Thanks @Krishna-kg732 for pointing this out..
@Kartikeya-trivedi can we follow the similar analogy if possible?

Kartikeya-trivedi:
Hi @abhijeet-dhumal , @Krishna-kg732 

sdk's file still says "Dependabot cannot bump curl-installed binaries — update version + checksum manually" , the exact manual chore this PR eliminates.
sdk is still pinned to 2.3.8, which is now stale (2.4.0 shipped weeks ago) that is the living proof of the drift problem PR [#90](https://github.com/kubeflow/mcp-server/pull/90) solves. sdk doesn't have the answer; it has the disease.
Their workflow list has no bump workflow at all. And the requested rename is impossible anyway: osv-scanner.yaml already exists in the repo
Bonus: sdk's own repo splits OSV concerns into separate files (osv-scanner.yaml + cleanup-overrides.yaml) so separate-file-per-concern literally is the sdk pattern. osv-scanner-bump.yaml follows it

Krishna-kg732:
hey @Kartikeya-trivedi , thanks for working on this 

/lgtm

cc @abhijeet-dhumal

Krishna-kg732:
looks like your branch is behind upstream , can you please rebase it for cleaner commit history

Kartikeya-trivedi:
Hi @Krishna-kg732 , I have rebased it as requested

jaiakash (approved):

abhijeet-dhumal:
/lgtm
/approve
Thanks @Kartikeya-trivedi ! 🚀

PR #104, opened 2026-07-21
fix(core): derive instruction section order dynamically from client modules

Krishna-kg732:
## Description

`_sections_for_persona()` hardcoded the instruction section order in
`server.py`, so every new client module required a core edit to have its
sections appear at all.

- Each client module now exports `SECTION_ORDER`; `server.py` concatenates them in `CLIENT_MODULES` order
- Adding a client no longer touches `server.py`
- Deterministic fallback: a section from a module that exports no `SECTION_ORDER`, or a name missing from it, still surfaces, sorted, so the result never depends on set iteration order
- `trainer` exports its existing order, so behavior is unchanged for the current single client

Rebased onto current main. The earlier `cli_test.py` hunk was dropped because
main already fixed that test independently.

## Related Issue

No separate issue. Prerequisite for the optimizer client (#105), whose sections
were previously invisible because they were not in the hardcoded list.

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Manually tested

Full suite green (474 passed, 14 deselected). `test_section_order` now asserts
structural invariants rather than a hardcoded list, so it does not break each
time a client is added:

- No duplicate sections
- Every section belongs to some client's `SECTION_ORDER`
- Relative order within a client is preserved

A dedicated test covers the fallback path, asserting that sections from a module
without `SECTION_ORDER` come back sorted rather than in arbitrary order.

### DIFF
--- a/kubeflow_mcp/core/server.py
+++ b/kubeflow_mcp/core/server.py
@@ -264,7 +264,15 @@ def _sections_for_persona(persona: str) -> list[str]:
     allowed = get_allowed_tools(persona) or set(TOOL_TO_PHASE.keys())
     phases_used = {TOOL_TO_PHASE[t] for t in allowed if t in TOOL_TO_PHASE}
 
-    section_order = ["planning", "monitoring", "training", "platform"]
+    # Canonical ordering is assembled from each client module's SECTION_ORDER
+    # export, so adding a client never requires editing server.py.
+    preferred_order: list[str] = []
+    for module_path in CLIENT_MODULES.values():
+        try:
+            module = importlib.import_module(module_path)
+            preferred_order.extend(getattr(module, "SECTION_ORDER", []))
+        except ImportError:
+            continue
 
     sections_needed: set[str] = set()
     for module_path in CLIENT_MODULES.values():
@@ -278,7 +286,12 @@ def _sections_for_persona(persona: str) -> list[str]:
         except ImportError:
             continue
 
-    return [s for s in section_order if s in sections_needed]
+    ordered = [s for s in preferred_order if s in sections_needed]
+    # Deterministic fallback: a section from a module that exports no
+    # SECTION_ORDER (or a name missing from it) still appears, sorted, so the
+    # result never depends on set iteration order.
+    extra = sorted(sections_needed - set(preferred_order))
+    return ordered + extra
 
 
 def _build_server_instructions(
--- a/kubeflow_mcp/core/server.py
+++ b/kubeflow_mcp/core/server.py
@@ -259,6 +259,12 @@ def _derive_tier(full_text: str, tier: str) -> str:
     return f"Tools: {', '.join(dict.fromkeys(tools))}"
 
 
+# Ordering used for a client that maps phases to sections but exports no
+# SECTION_ORDER. Matches the list that was hardcoded here before clients
+# declared their own, so such a client keeps its previous ordering.
+_DEFAULT_SECTION_ORDER = ["planning", "monitoring", "training", "platform"]
+
+
 def _sections_for_persona(persona: str) -> list[str]:
     """Derive which instruction sections a persona needs from its tool access."""
     allowed = get_allowed_tools(persona) or set(TOOL_TO_PHASE.keys())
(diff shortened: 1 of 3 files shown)


kubeflow_mcp/trainer/api/architecture_test.py
```diff
@@ -170,10 +170,47 @@ def test_sections_for_platform_admin(self):
         sections = _sections_for_persona("platform-admin")
         assert "platform" in sections
 
-    def test_section_order(self):
+    def test_section_order_is_deterministic_and_follows_client_order(self):
```
Krishna-kg732, line 173: rename

abhijeet-dhumal: you are reviewing your own PRs ? 😆

Krishna-kg732: 😬

kubeflow_mcp/core/server.py
```diff
@@ -264,7 +264,16 @@ def _sections_for_persona(persona: str) -> list[str]:
     allowed = get_allowed_tools(persona) or set(TOOL_TO_PHASE.keys())
     phases_used = {TOOL_TO_PHASE[t] for t in allowed if t in TOOL_TO_PHASE}
 
-    section_order = ["planning", "monitoring", "training", "platform"]
+    # Build the canonical ordering dynamically from each client module's
+    # SECTION_ORDER export.  This means adding a new client never requires
+    # touching server.py — just export SECTION_ORDER from the client package.
+    preferred_order: list[str] = []
+    for module_path in CLIENT_MODULES.values():
```
abhijeet-dhumal, line 270: Worth dict.fromkeys(preferred_order) (or similar) so the return list stays unique.. WDYT?

kubeflow_mcp/core/server.py
```diff
@@ -264,7 +264,16 @@ def _sections_for_persona(persona: str) -> list[str]:
...
+        try:
+            module = importlib.import_module(module_path)
+            preferred_order.extend(getattr(module, "SECTION_ORDER", []))
```
abhijeet-dhumal, line 273: If a client maps phases to sections but forgets SECTION_ORDER, we silently fall back to alphabetical 'extra'. That changes instruction order vs the old hardcoded list.

abhijeet-dhumal: Either require SECTION_ORDER whenever PHASE_TO_SECTION is set, or keep a trainer-compatible default fallback so missing SECTION_ORDER doesn’t reorder.

abhijeet-dhumal (commented):
For first release v0.1 I’d park this ... only trainer ships first. 
Happy to take it once optimizer lands, or split the cli_test fix into its own tiny PR if that test is failing on main.

kubeflow_mcp/core/server.py
Krishna-kg732: 🤔 , makes sense

kubeflow_mcp/core/server.py
Krishna-kg732: okayy , ack , i will add a default fallback

abhijeet-dhumal:
Prereq for #105, makes sense now. 
One fix needed from prior review (dedup), otherwise LGTM.

kubeflow_mcp/core/server.py
```diff
@@ -278,7 +286,12 @@ def _sections_for_persona(persona: str) -> list[str]:
         except ImportError:
             continue
 
-    return [s for s in section_order if s in sections_needed]
+    ordered = [s for s in preferred_order if s in sections_needed]
```
abhijeet-dhumal, line 289:
```suggestion
    ordered = list(dict.fromkeys(s for s in preferred_order if s in sections_needed))
```

kubeflow_mcp/trainer/api/architecture_test.py
```diff
@@ -170,10 +170,64 @@ def test_sections_for_platform_admin(self):
...
+        """A client exporting no SECTION_ORDER must still surface its sections,
+        in a deterministic (sorted) position rather than set iteration order."""
+        from types import SimpleNamespace
+        from unittest.mock import patch
+
+        # Two unknown sections, deliberately given in non-alphabetical order.
+        fake = SimpleNamespace(PHASE_TO_SECTION={"planning": "zeta", "discovery": "alpha"})
+        with (
+            patch("kubeflow_mcp.core.server.CLIENT_MODULES", {"fake": "fake_mod"}),
+            patch("importlib.import_module", return_value=fake),
```
abhijeet-dhumal, line 225: Nit: patch "kubeflow_mcp.core.server.importlib.import_module" instead .. global patch can leak.

abhijeet-dhumal:
@Krishna-kg732 are you still working on this ?

Krishna-kg732:
done 🫡

ps: The PR title check ci is failing :/

abhijeet-dhumal:
Thanks @Krishna-kg732 !
/lgtm
/approve

abhijeet-dhumal:
> done 🫡
> 
> ps: The PR title check ci is failing :/

It's happening due to a race condition .. I will raise fix for it.. promptly!

PR #105, opened 2026-07-21
feat(optimizer): add Katib client with 17 MCP tools (KEP-34)

Krishna-kg732:
## Description

Implements the Katib optimizer client from KEP-34: 17 MCP tools for experiment
planning, discovery, monitoring, creation and lifecycle.

> [!NOTE]
> Stacked on #157. That PR is the first commit here and drops out of this diff
> once it merges. Review #157 first.

| Phase | Tools |
|---|---|
| Planning | `katib_pre_flight` |
| Discovery | `list_experiments`, `get_experiment`, `get_experiment_status`, `get_trial`, `get_successful_trials`, `list_suggestions` |
| Monitoring | `get_experiment_trials`, `get_best_trial`, `get_suggestion`, `wait_for_experiment`, `get_experiment_trial_logs`, `get_experiment_events` |
| Optimization | `create_hpo_experiment`, `create_experiment_from_spec` |
| Lifecycle | `delete_experiment`, `update_experiment` |

**Key decisions**

- `OptimizerClient.optimize()` is unusable from MCP, so both creation tools build the `V1beta1Experiment` CR directly. This also reaches all 8 Katib algorithms instead of the 2 the SDK's typed classes express.
- `update_experiment` pauses via `spec.parallelTrialCount = 0`, not `spec.resumePolicy`, which governs restart after completion rather than a mid-run pause.
- `check_namespace_allowed()` gains an optional `resolver` so optimizer tools validate the namespace they actually target. The trainer path is unchanged.
- `get_experiment_status` reads the Experiment CR rather than `get_job()`, so it agrees with `get_experiment` and is cheaper.

<details>
<summary>Why <code>optimize()</code> cannot be used</summary>

It requires `TrainJobTemplate.trainer` to be a `CustomTrainer` holding a live
Python callable, which cannot cross the MCP/JSON boundary. It also generates its
own Experiment name and sets no labels, so a caller-supplied name and the MCP
ownership label are both unreachable. `SDK_COMPATIBILITY` records this under
`uncovered_methods`, and SDK contract tests fail if any constraint is relaxed.
</details>

<details>
<summary>Behaviour change in <code>get_experiment_status</code></summary>

It previously derived trial counts from TrainJob statuses while `get_experiment`
read the CR, so the two reported different numbers under the same field names.
It now also returns `early_stopped_trials`, `pending_trials` and
`killed_trials`. Counts differ from before whenever trials were early stopped.
The previous numbers were wrong.
</details>

<details>
<summary>Correctness fixes and API details</summary>

- `create_hpo_experiment` derives `primaryContainerName` from the trial template instead of hardcoding `node`, which silently collected no metrics from a plain `batch/v1` Job. Overridable via `primary_container_name`.
- Search space accepts `int` (with optional `step`). The SDK's `Search` emits only `double` and `categorical`, so an ordered discrete range could otherwise only be `choices`, which algorithms treat as unordered.
- `get_trial(name, trial)` takes the experiment as `name`, matching every other tool and `get_training_logs(name, step)`.
- `delete_experiment` and `update_experiment` report a missing experiment as `RESOURCE_NOT_FOUND` instead of blaming ownership.
- A create rejected by an unreachable Katib webhook now hints at `katib_pre_flight()`.
- `pyproject.toml` excludes `kubernetes` 36.0.0 and 36.0.1, which send no `Authorization` header at all (fixed upstream in 36.0.2, kubernetes-client/python#2595).
</details>

**Not included:** live validation against a running experiment. Katib on our test
cluster is missing its ClusterRoles, so the controller never becomes Ready and
every create is rejected by its own admission webhook.

## Related Issue

Implements KEP-34 (`docs/proposals/KEP-34.md`).

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Manually tested

- 742 passed, 14 deselected. Optimizer coverage 95 percent, every file above the 80 percent KEP target.
- Architecture tests bind advertised algorithms, search types and limits to the implementation, so the docs cannot drift from the code.
- The namespace-allowlist sweep is derived from `TOOLS`, so a new tool cannot skip coverage.
- Exercised live: pre-flight, all read-only tools, every not-found path, and create validation and preview. Pre-flight correctly reported the broken controller rather than a false green.

### DIFF
--- a/kubeflow_mcp/optimizer/api/lifecycle.py
+++ b/kubeflow_mcp/optimizer/api/lifecycle.py
@@ -220,7 +220,9 @@ def update_experiment(
                 "metadata": {"annotations": annotation_patch},
                 "spec": {"parallelTrialCount": target},
             },
-            _request_timeout=mcp_utils.K8S_TIMEOUT,
+            # Patching an Experiment is admitted by the same webhooks as a
+            # create, so it gets the write budget rather than the read timeout.
+            _request_timeout=mcp_utils.K8S_WRITE_TIMEOUT,
         )
 
         past = "suspended" if action == "suspend" else "resumed"
--- a/kubeflow_mcp/optimizer/api/lifecycle.py
+++ b/kubeflow_mcp/optimizer/api/lifecycle.py
@@ -47,6 +47,41 @@ ACTIONS = ("suspend", "resume")
 PARALLELISM_ANNOTATION = "kubeflow-mcp/pre-suspend-parallel-trial-count"
 DEFAULT_PARALLELISM = 1
 
+# (parallelTrialCount to set, annotation patch to apply)
+ParallelismPatch = tuple[int, dict[str, str | None]]
+
+
+def _suspend_patch(current_parallelism: int | None) -> ParallelismPatch:
+    """Pause by zeroing concurrency, remembering what to restore later."""
+    remembered = str(current_parallelism or DEFAULT_PARALLELISM)
+    return 0, {PARALLELISM_ANNOTATION: remembered}
+
+
+def _resume_patch(annotations: dict[str, str]) -> ParallelismPatch:
+    """Restore the remembered concurrency, clearing the stash as we go.
+
+    Never restores 0: a missing or corrupt annotation would otherwise leave the
+    experiment paused while reporting that it resumed.
+    """
+    stashed = annotations.get(PARALLELISM_ANNOTATION)
+    target = int(stashed) if stashed and stashed.isdigit() else DEFAULT_PARALLELISM
+    return max(target, DEFAULT_PARALLELISM), {PARALLELISM_ANNOTATION: None}
+
+
+def _update_response(
+    name: str, namespace: str, action: str, parallelism: int, outcome: str
+) -> dict[str, Any]:
+    """Single response shape for every update_experiment outcome."""
+    return ToolResponse(
+        data={
+            "experiment": name,
+            "namespace": namespace,
+            "action": action,
+            "parallel_trial_count": parallelism,
+            "message": f"Experiment '{name}' {outcome}",
+        }
+    ).model_dump()
+
 
 def delete_experiment(
     name: str,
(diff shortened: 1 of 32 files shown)


abhijeet-dhumal:
Hey @Krishna-kg732 Can you rebase this work based on conventions documented here : https://github.com/kubeflow/mcp-server/blob/main/docs/CONVENTIONS.md

Krishna-kg732:
done 🫡

abhijeet-dhumal:
@Krishna-kg732  Quick update on this work : This PR builds against SDK 0.4.1, which already ships OptimizerClient so the tools themselves are functional. However, to land this cleanly we should first complete [#195](https://github.com/kubeflow/mcp-server/issues/195) (bump kubeflow to 0.5.0). The 0.5.0 SDK brings kubeflow-trainer-api==2.3.0 and aligns all clients to the same wire format, avoiding a mid-milestone dep mismatch between trainer and optimizer tools. (targeted for next Kubeflow-MCP major 0.2.0 release)
cc: @andreyvelich @jaiakash

kubeflow_mcp/optimizer/api/lifecycle.py
```diff
@@ -0,0 +1,233 @@
...
+                "message": f"Experiment '{name}' deleted",
+            }
+        ).model_dump()
+
+    except Exception as e:
+        logger.warning("delete_experiment(%s) failed: %s", name, e, exc_info=True)
+        return experiment_error(e, name).model_dump()
+
+
+def update_experiment(
```
abhijeet-dhumal, line 150: This mutates an experiment without a confirmed preview gate, violating the repository’s mandatory confirmation semantics. Please add confirmed: bool = False and return a preview until confirmed.

kubeflow_mcp/optimizer/api/optimization.py
```diff
@@ -0,0 +1,461 @@
...
+    if not name:
+        return ToolError(
+            error="spec.metadata.name is required",
+            error_code=ErrorCode.VALIDATION_ERROR,
+        ).model_dump()
+    name_err = validate_k8s_name(name)
+    if name_err is not None:
+        return name_err.model_dump()
+
+    ns_err = check_optimizer_namespace(namespace)
```
abhijeet-dhumal, line 424: Namespace validation uses the default namespace when namespace=None, but the raw spec can target another namespace. Resolve the final target namespace from the spec first, then validate that exact namespace against policy.

PR #107, opened 2026-07-21
fix(core): pin kubeflow to exact version 0.4.0

### LINKED ISSUE #106: fix(core): Pin all pyproject.toml dependencies to exact versions — kubeflow==0.4.0 conflicts with fine_tune() fix
## Request

Exact-pin every dependency in `pyproject.toml` (`dependencies`,
`optional-dependencies`, `dependency-groups`) instead of using `>=`
ranges, for reproducible installs. In particular, `kubeflow>=0.4.0`
(pyproject.toml:22, and the `hub`/`spark`/`all` extras at lines 27-29)
should become `kubeflow==0.4.0`.

## Compatibility risk to resolve first

The environment currently resolves and locks `kubeflow` at **0.4.1**
(`uv.lock`), not 0.4.0, because `>=0.4.0` lets the resolver pick the
latest compatible release. Pinning down to exactly `0.4.0` is a
**downgrade** from what's installed and tested today.

This matters because of #<fine-tune-issue-number> / the fix in
`fix/remove-stale-hf-dataset-trainerargs-workaround`
(commit e49d818): that fix removed a workaround for a bug where the SDK
built an invalid `dataset.data_dir` for top-level HuggingFace dataset
URIs (`hf://org/name`). The workaround was removed because SDK 0.4.1
fixed the underlying bug natively
(`kubeflow/trainer/backends/kubernetes/utils.py`) and started rejecting
the old `TrainerArgs` override on `BuiltinTrainer`
(`kubeflow/trainer/options/kubernetes.py`).

If `kubeflow` is pinned back to `0.4.0`:
- It's unconfirmed whether 0.4.0 has the same `TrainerArgs`/`BuiltinTrainer`
  guard. If it doesn't, the removed workaround's absence may reintroduce
  the original `dataset.data_dir` bug for plain `hf://org/name` datasets
  (the common case) — i.e. `fine_tune()` breaks again, differently.
- Needs to be verified against 0.4.0 directly before merging the pin.

## Proposed scope

- [ ] Verify `kubeflow==0.4.0` behavior for `fine_tune()` with a
      top-level HF dataset (e.g. `hf://tatsu-lab/alpaca`) — does the
      original `dataset.data_dir` bug reproduce?
- [ ] If it does: either pin `kubeflow==0.4.1` instead, or restore a
      version-gated workaround.
- [ ] If it doesn't: proceed with `kubeflow==0.4.0`.
- [ ] Exact-pin remaining dependencies (`fastmcp`, `pydantic`, `pyyaml`,
      `click`, `huggingface-hub`, and the `dev`/`docs`/`otel` groups) to
      their currently locked `uv.lock` versions.
- [ ] Run `uv lock` and confirm `uv run pytest` / `uv run ruff check`
      pass against the pinned set.

myselfkunal:
## Summary

Fixes #106

Pins `kubeflow` to the exact version `0.4.0` in `pyproject.toml`, replacing the `>=0.4.0` range that let the resolver drift to `0.4.1` (as confirmed by `uv.lock` locking `0.4.1` instead of the intended `0.4.0`).

## Changes

- `dependencies`: `kubeflow>=0.4.0` → `kubeflow==0.4.0`
- `optional-dependencies.hub`: `kubeflow[hub]>=0.4.0` → `kubeflow[hub]==0.4.0`
- `optional-dependencies.spark`: `kubeflow[spark]>=0.4.0` → `kubeflow[spark]==0.4.0`
- `optional-dependencies.all`: both extras pinned to `==0.4.0`

## Compatibility check

Per the issue's checklist, verified `fine_tune()` behavior with a top-level HF dataset (`hf://tatsu-lab/alpaca`) against `kubeflow==0.4.0` directly to confirm the `dataset.data_dir` bug (originally worked around in `fix/remove-stale-hf-dataset-trainargs-workaround`, commit `e49d818`) does not reintroduce itself at this pinned version.

## Testing

```bash
uv lock
uv run pytest
uv run ruff check
```

All passing against the pinned dependency set.

## Notes

The remaining dependencies (`fastmcp`, `pydantic`, `pyyaml`, `click`, `huggingface-hub`, and the `dev`/`docs`/`otel` groups) were left as-is in this PR — happy to follow up with exact pins for those against `uv.lock` if maintainers want it scoped into this same PR instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,14 +19,14 @@ dependencies = [
     "pydantic>=2.0",
     "pyyaml>=6.0",
     "click>=8.0",
-    "kubeflow>=0.4.0",
+    "kubeflow==0.4.0",
     "huggingface-hub>=0.20",
 ]
 
 [project.optional-dependencies]
-hub = ["kubeflow[hub]>=0.4.0"]
-spark = ["kubeflow[spark]>=0.4.0"]
-all = ["kubeflow[hub]>=0.4.0", "kubeflow[spark]>=0.4.0"]
+hub = ["kubeflow[hub]==0.4.0"]
+spark = ["kubeflow[spark]==0.4.0"]
+all = ["kubeflow[hub]==0.4.0", "kubeflow[spark]==0.4.0"]
 
 [dependency-groups]
 dev = [
--- a/uv.lock
+++ b/uv.lock
@@ -48,7 +48,7 @@ resolution-markers = [
     "python_full_version < '3.11' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "caio", marker = "python_full_version < '3.11'" },
+    { name = "caio" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/67/e2/d7cb819de8df6b5c1968a2756c3cb4122d4fa2b8fc768b53b7c9e5edb646/aiofile-3.9.0.tar.gz", hash = "sha256:e5ad718bb148b265b6df1b3752c4d1d83024b93da9bd599df74b9d9ffcf7919b", size = 17943, upload-time = "2024-10-08T10:39:35.846Z" }
 wheels = [
@@ -70,7 +70,7 @@ resolution-markers = [
     "python_full_version == '3.11.*' and sys_platform != 'win32'",
 ]
 dependencies = [
-    { name = "caio", marker = "python_full_version >= '3.11'" },
+    { name = "caio" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/48/41/2fea7e193e061ce54eacc3b7bc0e6a99e4fcff43c78cf0a76dd781ed8334/aiofile-3.11.1.tar.gz", hash = "sha256:1f91912c6643d2a4e49ca4ae3514f0bf3867ce948a36d99a6411b8f4755f4cf9", size = 19342, upload-time = "2026-05-16T08:18:33.538Z" }
 wheels = [
@@ -1350,7 +1350,7 @@ name = "importlib-metadata"
 version = "9.0.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "zipp", marker = "python_full_version < '3.12'" },
+    { name = "zipp" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/a9/01/15bb152d77b21318514a96f43af312635eb2500c96b55398d020c93d86ea/importlib_metadata-9.0.0.tar.gz", hash = "sha256:a4f57ab599e6a2e3016d7595cfd72eb4661a5106e787a95bcc90c7105b831efc", size = 56405, upload-time = "2026-03-20T06:42:56.999Z" }
 wheels = [
@@ -1507,7 +1507,7 @@ wheels = [
 
 [[package]]
 name = "kubeflow"
-version = "0.4.1"
+version = "0.4.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "kubeflow-katib-api" },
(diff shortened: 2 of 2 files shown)


Krishna-kg732 (changes requested):
Hey @myselfkunal  ,Thanks for picking this up 

few nits , please take a look

pyproject.toml
```diff
@@ -19,14 +19,14 @@ dependencies = [
...
     "huggingface-hub>=0.20",
 ]
 
 [project.optional-dependencies]
-hub = ["kubeflow[hub]>=0.4.0"]
-spark = ["kubeflow[spark]>=0.4.0"]
-all = ["kubeflow[hub]>=0.4.0", "kubeflow[spark]>=0.4.0"]
+hub = ["kubeflow[hub]==0.4.0"]
+spark = ["kubeflow[spark]==0.4.0"]
+all = ["kubeflow[hub]==0.4.0", "kubeflow[spark]==0.4.0"]
```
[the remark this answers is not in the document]
Krishna-kg732: cc : @abhijeet-dhumal

myselfkunal:
> Hey @myselfkunal ,Thanks for picking this up
> 
> few nits , please take a look

Hey @Krishna-kg732 , pushed an update (d72cc10) that regenerates uv.lock and exact-pins the remaining dev/docs/otel deps too , should cover the full scope now. Let me know if anything's still missing!

pyproject.toml
```diff
@@ -15,36 +15,39 @@ classifiers = [
...
 dependencies = [
-    "fastmcp>=2.0.0",
-    "pydantic>=2.0",
-    "pyyaml>=6.0",
-    "click>=8.0",
-    "kubeflow>=0.4.0",
-    "huggingface-hub>=0.20",
+    "fastmcp==3.4.4",
+    "pydantic==2.13.4",
+    "pyyaml==6.0.3",
```
Krishna-kg732: Hey @myselfkunal , the issue and your PR states to only pin kubeflow to 0.4.0 , it looks. like you have also pinned other dependencies to their exact versions , can you please take a look and rerun the tests

Krishna-kg732 (changes requested):

pyproject.toml
myselfkunal: Done @Krishna-kg732 , reverted the other deps, now only kubeflow>=0.4.0 → ==0.4.0 is pinned
(main deps + hub/spark/all extras), plus the regenerated uv.lock to match.
make verify and make test-python both pass.

abhijeet-dhumal:
Hey @myselfkunal, can you resolve Precommit PR check failure first ? 
`uv run pre-commit run --all-files `

abhijeet-dhumal:
@myselfkunal this PR doesn't show any files anymore 🤔

abhijeet-dhumal:
@myselfkunal we are so close.. just need to sign off commit to fix PR check failure 
`git commit -s -m ...`

pyproject.toml
```diff
@@ -19,14 +19,14 @@ dependencies = [
...
-    "kubeflow>=0.4.0",
+    "kubeflow==0.4.0",
     "huggingface-hub>=0.20",
 ]
 
 [project.optional-dependencies]
-hub = ["kubeflow[hub]>=0.4.0"]
-spark = ["kubeflow[spark]>=0.4.0"]
-all = ["kubeflow[hub]>=0.4.0", "kubeflow[spark]>=0.4.0"]
+hub = ["kubeflow[hub]==0.4.0"]
```
abhijeet-dhumal, line 27: nit: uv.lock has a lot of unrelated marker churn,.. if easy, regenerate so the lock diff is just kubeflow 0.4.1 to 0.4.0.

abhijeet-dhumal (approved):
/lgtm
/approve
Thanks @myselfkunal for your contribution 🚀

PR #28, opened 2026-05-28
feat: create initial release workflow for automated pypi publishing

Krishna-kg732:
## Automate the release pipeline for kubeflow-mcp (SDK-aligned)

Replaces the earlier manual-publish design with a version-file-driven, automated pipeline modeled on the [kubeflow/sdk release workflow](https://github.com/kubeflow/sdk/blob/main/RELEASE.md). Merging a version bump now drives branch creation, tagging, build, and publish automatically instead of requiring a manually published GitHub Release.

### What changed

- **Trigger model**: `push` to `main` / `release-*`, filtered on `kubeflow_mcp/__init__.py` changing, plus `workflow_dispatch` for a Test PyPI dry run only. The old `release: published` trigger, `force_prod`, and `dry_run` inputs are gone.
- **Single version source**: `pyproject.toml` now reads its version dynamically from `kubeflow_mcp/__init__.py` via `[tool.hatch.version]`. `scripts/check_version.py` is deleted since there is nothing left to keep in sync.
- **Automated branch/tag**: `prepare` creates or cherry-picks onto `release-X.Y`, `create-tag` tags the release. Both mirror the SDK's workflow.
- **`make release VERSION=X.Y.Z[rcN]`**: new Makefile target that bumps the version. (Changelog generation via git-cliff is intentionally left commented out pending #72, which adds it separately.)
- **Docker publish reconciled**: `docker-publish.yaml` is now a reusable `workflow_call` invoked from the release pipeline after tagging (tags pushed by `GITHUB_TOKEN` can't retrigger a tag-push workflow, so it's called directly). Also fixes a pre-existing bug where an RC build could get tagged `latest`.
- **Dispatch-branch guard**: a fail-fast step rejects `workflow_dispatch` from any branch other than `main`/`release-*`, since GitHub Actions silently ignores a `branches:` filter on that trigger. Addresses review feedback from @abhijeet-dhumal.
- **RELEASE.md rewritten** end to end for the automated flow.

### Environments required (unchanged)

- `release` and `test-pypi` GitHub environments with required reviewers.
- PyPI + Test PyPI trusted publishers for `kubeflow-mcp` (owner `kubeflow`, repo `mcp-server`, workflow `release.yaml`).

### Verified

- `make verify` and `make test-python` pass.
- Dry ran the full `workflow_dispatch` path end to end on a fork (prepare/build/Test PyPI publish all green).
- Guard logic checked against `main`, `release-0.1`, and a feature branch ref.

### Out of scope

[...]

PR #117, opened 2026-07-30
fix(ci): create GitHub release as draft before uploading assets

### LINKED ISSUE #115: bug: fix the release workflow for first rc
## Description

The release workflow will fail at the `github-release` job. It publishes the release directly (`draft: false`) while attaching `dist/*` assets — with immutable releases enabled, assets are locked once a release is published, so the asset upload fails.

As discussed in today's SDK WG call, can @Krishna-kg732 please fix the workflow to match the SDK approach:

- Create the release as **draft**, upload assets, then publish with `gh release edit "$VERSION" --draft=false`
- Pin `softprops/action-gh-release` to a commit SHA (v3), same as SDK

Reference: https://github.com/kubeflow/sdk/blob/main/.github/workflows/release.yaml

/assign @Krishna-kg732

Krishna-kg732:
## Description

The `github-release` job created the release with `draft: false` while attaching `dist/*`. With immutable releases enabled, a published release locks its assets, so the upload fails.

Fixes it the same way the SDK does:

- create the release as a draft, upload assets, then publish with `gh release edit "$VERSION" --draft=false`
- pin `softprops/action-gh-release` to a commit SHA (v3)

While here, brings the workflow to parity with [kubeflow/sdk](https://github.com/kubeflow/sdk/blob/main/.github/workflows/release.yaml): drops the Test PyPI job and the `workflow_dispatch` input (now `{}`), removes the per-job `if: github.event_name == 'push'` guards so a dispatch behaves like a version-bump push, and bumps `astral-sh/setup-uv` to v7. RELEASE.md updated to match.

Two guards are kept on purpose, both noted in the workflow:

- dispatch is restricted to `main`/`release-*` (per review on #28, and it matters more now that a dispatch cuts a real release)
- the version must be `X.Y.Z[rcN]`, so an unrelated edit to `__init__.py` while the version is the `0.1.0-dev` placeholder can't cut a bogus release

## Related Issue

Fixes #115

## Checklist

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests

`make verify` (incl. `check-yaml`) and `make test-python` (218 passed) pass locally. Workflow YAML parses with the expected job graph, and the pinned SHA `3d0d9888` is confirmed to be exactly what the `v3` tag resolves to.

The publish path itself can't be exercised before merge, since it needs the `release` environment and OIDC on the canonical repo.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,15 +4,10 @@ on:
   push:
     branches:
       - main
-      - 'release-*'
+      - "release-*"
     paths:
-      - 'kubeflow_mcp/__init__.py'
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'Expected version for the Test PyPI dry run (must match kubeflow_mcp/__init__.py)'
-        required: false
-        type: string
+      - "kubeflow_mcp/__init__.py"
+  workflow_dispatch: {}
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


Krishna-kg732:
Fork Test: https://github.com/Krishna-kg732/KF-mcp-server/actions/runs/30531476971
Test-PyPI: https://test.pypi.org/project/kubeflow-mcp-krishna-kg732/

PR #142, opened 2026-08-04
fix: handle non-mapping YAML config

### LINKED ISSUE #141: fix: handle non-mapping YAML configuration safely
## Problem

`_load_yaml_config()` in `kubeflow_mcp/core/config.py` returns any value produced by `yaml.safe_load()`.

The configuration loader later assumes the root YAML value is a dictionary and calls `.get()` on it:

```python
server_file = file_config.get("server", {})
```

If a user provides syntactically valid YAML whose root value is not a mapping, for example:

```yaml
- trainer
- optimizer
```

`yaml.safe_load()` returns a list. The subsequent `.get()` call raises an `AttributeError`, causing configuration loading to fail unexpectedly.

## Proposed fix

Validate that the root YAML value is a mapping before returning it from `_load_yaml_config()`.

For a non-mapping root value, log a clear warning and fall back to an empty configuration, matching the existing behavior for malformed YAML files.

## Testing

Add a focused test in `kubeflow_mcp/core/config_test.py` that writes a valid YAML list as the root value and verifies that `load_config()` safely falls back to the built-in defaults.

## Acceptance criteria

- A YAML configuration file with a non-mapping root value does not raise `AttributeError`.
- The server falls back safely to default configuration values.
- A clear warning is logged for invalid root configuration structure.
- A unit test covers this behavior.

pratik-naik003:
## Description

Handles YAML configuration files whose root value is not a mapping.

Previously, valid YAML with a list or scalar root could reach `load_config()` and raise `AttributeError` when the loader called `.get()` on a non-dictionary value.

The loader now logs a warning and safely falls back to default configuration values. A focused regression test covers a YAML list root.

## Related Issue

Fixes #141

## Checklist

- [ ] I have read the [CONTRIBUTING](./CONTRIBUTING.md) guide
- [ ] Tests pass locally (`make test-python`)
- [ ] Linting passes (`make verify`)

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

Ran:

```bash
uv run pytest -q kubeflow_mcp/core/config_test.py
uv run ruff check kubeflow_mcp/core/config.py kubeflow_mcp/core/config_test.py
uv run ruff format --check kubeflow_mcp/core/config.py kubeflow_mcp/core/config_test.py
```

All 10 configuration tests passed, and lint/format checks passed.

### DIFF
--- a/kubeflow_mcp/core/config.py
+++ b/kubeflow_mcp/core/config.py
@@ -191,7 +191,14 @@ def _load_yaml_config(path: Path) -> dict[str, Any]:
 
         with open(path) as f:
             data = yaml.safe_load(f)
-            return data if data else {}
+            if data is None:
+                return {}
+            if not isinstance(data, dict):
+                logger.warning(
+                    "Config root must be a mapping; ignoring %s (got %s)", path, type(data).__name__
+                )
+                return {}
+            return data
     except ImportError:
         logger.warning("PyYAML not installed, skipping config file")
         return {}
--- a/kubeflow_mcp/core/config_test.py
+++ b/kubeflow_mcp/core/config_test.py
@@ -108,6 +108,16 @@ def test_load_config_env_overrides_file(tmp_path):
     assert cfg.server.persona == "platform-admin"
 
 
+def test_load_config_ignores_non_mapping_yaml(tmp_path, caplog):
+    config_file = tmp_path / "config.yaml"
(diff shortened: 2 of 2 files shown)


abhijeet-dhumal:
/lgtm
/approve
Thanks @pratik-naik003, great catch !

PR #156, opened 2026-08-11
chore: kubeflow MCP Server release 0.1.1

abhijeet-dhumal:
## Description

Re-release of 0.1.0 to work around a PyPI filename reuse restriction.
The 0.1.0 wheel was previously uploaded and deleted from PyPI, permanently
blocking that filename. This bumps to 0.1.1 with identical content.
cc: @andreyvelich

## Related Issue

Fixes the PyPI publish failure in https://github.com/kubeflow/mcp-server/actions/runs/31501365444


andreyvelich (commented):
/lgtm
/approve

PR #158, opened 2026-08-12
fix(trainer): report a missing TrainJob as not found instead of unowned

Krishna-kg732:
## Description

- `is_mcp_managed` returned `False` for a TrainJob that does not exist, the same value it returns for one that exists without the MCP ownership label.
- `delete_training_job` and `update_training_job` gate on ownership before existence, so a mistyped name was reported as `VALIDATION_ERROR: "was not created by MCP"` with a permissions hint, instead of not found.
- The same missing job answered differently depending on the caller: `platform-admin` skips the gate and correctly got `RESOURCE_NOT_FOUND`, while every other persona got the misleading error.
- Replaced with `get_trainer_ownership`, returning `managed`, `unmanaged` or `missing`. Missing now maps to `RESOURCE_NOT_FOUND`, and the ownership message is reserved for jobs that actually exist.
- Both call sites now share one `_require_mcp_ownership` helper, removing duplicated error construction that had already drifted between the two tools.

No behaviour change for `managed`, `unmanaged` or API-failure outcomes.

## Related Issue

No existing issue. Found while exercising the Katib optimizer client against a live cluster.

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

- Added `test_missing_job_end_to_end_from_a_real_404`, which mocks only the Kubernetes API so the real ownership helper runs. The other ownership tests stub the helper, so they pin the gate's branching but cannot catch this class of regression.
- Verified the coverage actually bites: reintroducing the bug fails 3 tests, and all pass with the fix.
- Reproduced the original behaviour first by driving both tools with a mocked 404 across all three personas.
- Fills two existing `TODO(test)` markers in `trainer/api/lifecycle_test.py`.
- `make test-python`: 485 passed. `make verify`: all checks passed.

/kind bug
/area trainer

### DIFF
--- a/kubeflow_mcp/trainer/api/lifecycle_test.py
+++ b/kubeflow_mcp/trainer/api/lifecycle_test.py
@@ -14,12 +14,13 @@
 
 """Tests for trainer/api/lifecycle.py — delete, suspend, resume.
 
-Covers input validation. K8s API interaction tests require mocking and are
-marked as TODOs.
+Covers input validation and the MCP ownership gate. The remaining K8s API
+interaction tests are marked as TODOs.
 """
 
 from __future__ import annotations
 
+from contextlib import contextmanager
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -29,6 +30,7 @@ from tests.common import (
     FAILED,
     PERMISSION_DENIED,
     RESOURCE_NOT_FOUND,
+    SDK_ERROR,
     VALIDATION_ERROR,
     TestCase,
     assert_test_case,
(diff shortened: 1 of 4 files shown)


abhijeet-dhumal:
Thanks @Krishna-kg732, Nice catch and a clean fix,.. I guess the persona-dependent answer for a mistyped name was actively misleading agents. This is exactly what we needed 🎉

kubeflow_mcp/trainer/api/lifecycle_test.py
```diff
@@ -49,9 +73,117 @@ def test_delete_training_job_validation(test_case):
...
+
+
+@pytest.mark.parametrize(
+    ("tool", "call", "attr"),
+    [
+        ("delete", lambda: delete_training_job("external", confirmed=True), "delete_job"),
+        ("update", lambda: update_training_job("external", "suspend"), "delete_job"),
+    ],
+)
+def test_non_admin_blocked_from_mutating_external_job(tool, call, attr):
```
abhijeet-dhumal, line 118: Nit: the update row still passes attr="delete_job", but update_training_job never calls client.delete_job..  it just patches via the custom-objects API. The patch assert below already covers it..  
either drop attr for the update case or only assert delete_job on the delete row.

kubeflow_mcp/trainer/api/lifecycle_test.py
```diff
@@ -49,9 +73,117 @@ def test_delete_training_job_validation(test_case):
...
+        patch(f"{_UTILS}.get_trainer_effective_namespace", return_value="default"),
+        patch(f"{_UTILS}.get_custom_objects_api", return_value=k8s),
+        patch(f"{_UTILS}.get_trainer_custom_objects_api", return_value=k8s),
+        patch(f"{_UTILS}.get_trainer_client_for_namespace", return_value=client),
+        patch(f"{_LIFECYCLE}.get_effective_persona", return_value="data-scientist"),
+        patch(f"{_LIFECYCLE}.check_namespace_allowed", return_value=None),
+    ):
+        result = call()
+
+    assert result["error_code"] == "RESOURCE_NOT_FOUND"
```
abhijeet-dhumal, line 177: Nit: prefer RESOURCE_NOT_FOUND / SDK_ERROR from tests.common (same as VALIDATION_ERROR already imported) instead of raw strings.

abhijeet-dhumal:
@Krishna-kg732 are you still working on this ? Did you get time to resolve nit picks raised ?

abhijeet-dhumal:
Thanks @Krishna-kg732, Great catch !
/lgtm
/approve

PR #159, opened 2026-08-13
fix(core): redact sensitive structured fields

pulkit-999:
Fix : 

- Add recursive _redact_dict() that replaces sensitive key values (password, token, secret, authorization, credential, api_key) with "***" in dictionaries and lists.
- Apply to request_context and all extra structured fields (parameters, audit, tool, etc.) in StructuredFormatter.
- Add missing unit test kubeflow_mcp/core/logging_test.py to validate redaction.

Issue:

- StructuredFormatter currently logs context and extra fields without redaction, leaking credentials/tokens to production log aggregators. This closes that gap.

Testing

- New unit test covers nested dicts, lists, and mixed sensitive/non‑sensitive fields.

All existing tests pass (make test-python & make verify).

### DIFF
--- a/kubeflow_mcp/core/logging.py
+++ b/kubeflow_mcp/core/logging.py
@@ -24,6 +24,8 @@ from contextvars import ContextVar
 from datetime import datetime, timezone
 from typing import Any
 
+from kubeflow_mcp.core.security import mask_sensitive_data
+
 correlation_id: ContextVar[str] = ContextVar("correlation_id", default="")
 request_context: ContextVar[dict[str, Any] | None] = ContextVar("request_context", default=None)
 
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
Thanks @pulkit-999 , great catch !
One pick here .. : please don’t fork a second redaction policy. Reuse `mask_sensitive_data` from `core/security.py` for dict keys (same rules as audit + OTel), and optionally keep `_REDACT_PATTERNS` on string leaves / the log message. The current `"token" in key` match redacts `tokenizer`, which `mask_sensitive_data` deliberately does not.
Happy to /lgtm once that’s aligned.

pulkit-999:
> Thanks @pulkit-999 , great catch ! One pick here .. : please don’t fork a second redaction policy. Reuse `mask_sensitive_data` from `core/security.py` for dict keys (same rules as audit + OTel), and optionally keep `_REDACT_PATTERNS` on string leaves / the log message. The current `"token" in key` match redacts `tokenizer`, which `mask_sensitive_data` deliberately does not. Happy to /lgtm once that’s aligned.

Have implemented the required changes now👍

abhijeet-dhumal:
Thanks @pulkit-999, great catch !
/lgtm
/approve

abhijeet-dhumal:
Thanks @pulkit-999, great catch !
/lgtm
/approve

PR #160, opened 2026-08-15
feat(examples): add manifests for in-cluster deployment

szymka117:
## Description

Adds a self-contained Kubernetes example so the MCP Server can be deployed in a
cluster and reached over HTTP by an MCP client. A single `kubectl apply` creates the
namespace, ServiceAccount, least-privilege RBAC, auth Secret, Deployment and Service,
with liveness and readiness probes wired to `/health` and `/ready`.

## Related Issue

Related to kubeflow/mcp-server#161

## Checklist

- [x] Tests pass locally (`make test-python`) (failing for current main)

## Testing

Deployed to a kind cluster running Kubeflow Trainer v2.1.0, then driven end to end from
an MCP client (Claude Code) connected over `kubectl port-forward`.

**`pre_flight`** — environment compatible:

**`run_custom_training`** — the two-phase confirmation gate behaved as designed: the
first call with `confirmed=False` returned a clean preview with no safety warnings, and
the job was only submitted on the follow-up call with `confirmed=True`.

**Result** — TrainJob `two-moons-mlp` reached `Complete`, with logs retrieved through
the server:

### DIFF
--- a/examples/kubernetes/README.md
+++ b/examples/kubernetes/README.md
@@ -1,0 +1,161 @@
+# Deploying Kubeflow MCP Server in a cluster
+
+A minimal, self-contained example for running the MCP Server inside Kubernetes so
+that agents can reach it over HTTP and drive
+[Kubeflow Trainer](https://github.com/kubeflow/trainer) in your own namespace.
+
+It is intentionally small: one Deployment, one ClusterIP Service, a ServiceAccount
+and least-privilege RBAC. No Ingress, no OIDC, no Helm chart.
+
+## Prerequisites
+
+- A Kubernetes cluster with Kubeflow Trainer installed
+- A namespace to run in. The example uses `kubeflow-user-example-com`, the Profile
+  namespace created by a standard Kubeflow installation
+- `kubectl` pointed at that cluster
+
+## Quick start
+
+```bash
+NAMESPACE=kubeflow-user-example-com
+
+# Generate the bearer token and deploy everything in one step. Piping keeps the
+# token out of manifests.yaml, so no credential is ever written to disk.
+sed "s/REPLACE_ME/$(openssl rand -hex 32)/" manifests.yaml | kubectl apply -f -
+
+kubectl rollout status deploy/kubeflow-mcp -n "$NAMESPACE"
+```
+
+Read the token back when configuring a client:
+
+```bash
+kubectl get secret kubeflow-mcp-auth -n "$NAMESPACE" -o jsonpath='{.data.token}' | base64 -d
+```
+
+To rotate it later, re-run the command above and restart the pod so the new value
+is picked up:
+
+```bash
+kubectl rollout restart deploy/kubeflow-mcp -n "$NAMESPACE"
+```
+
+If you prefer to manage the Secret separately — for example with an external secret
+store — delete the `Secret` document from `manifests.yaml` and create it
+imperatively instead. Leaving the document in place while doing this would reset
+the token to `REPLACE_ME` on the next apply:
+
+```bash
+kubectl create secret generic kubeflow-mcp-auth -n "$NAMESPACE" \
+  --from-literal=token="$(openssl rand -hex 32)"
+```
+
+## What gets created
+
+| Resource | Purpose |
+|---|---|
+| `Namespace/kubeflow-user-example-com` | Already present on a standard Kubeflow install, where the Profile controller owns it — applying is then a no-op |
+| `ServiceAccount/kubeflow-mcp` | Identity the server uses against the Kubernetes API |
+| `ClusterRole/kubeflow-mcp-read` | Read-only: `ClusterTrainingRuntime`, nodes, namespaces, CRDs |
+| `Role/kubeflow-mcp-trainjobs` | Full TrainJob lifecycle, in this namespace only |
+| `Role/kubeflow-mcp-trainer-version` | Read of the single `kubeflow-trainer-public` ConfigMap in `kubeflow-system`, so the SDK can report the Trainer control-plane version |
+| `Secret/kubeflow-mcp-auth` | Bearer token for the HTTP transport |
+| `Deployment/kubeflow-mcp` | The server, HTTP transport on port 8000 |
+| `Service/kubeflow-mcp` | ClusterIP, reachable at `kubeflow-mcp:8000` in-namespace |
+
+The server can create and delete TrainJobs in its own namespace, and can only read
+anything outside it.
+
+## Verify
+
+From inside the cluster:
+
+```bash
+kubectl run mcp-check -n "$NAMESPACE" --rm -it --restart=Never --image=curlimages/curl -- \
+  curl -s http://kubeflow-mcp:8000/ready
+```
+
+`/health` and `/ready` are served without authentication so the kubelet can probe
+them; every other endpoint requires the bearer token.
+
+To reach it from your workstation:
+
+```bash
+kubectl port-forward -n "$NAMESPACE" svc/kubeflow-mcp 8000:8000
+```
+
+Then point an MCP client at `http://localhost:8000/mcp` with
+`Authorization: Bearer <token>`.
+
+`KUBEFLOW_MCP_ALLOWED_HOSTS` lists `localhost` and `127.0.0.1` alongside the Service
+names for this reason. Setting the variable replaces the built-in loopback defaults
+rather than adding to them, so dropping those two entries makes port-forwarded
+clients fail with HTTP 421 while `/health` and `/ready` keep working.
+
+## Deploying into a different namespace
+
+The server is deployed into the user's Profile namespace on purpose: in-cluster the
+Kubeflow SDK derives its default namespace from the pod's ServiceAccount namespace.
+Running it elsewhere makes every tool default to a namespace that has no TrainJobs
+in it.
+
+To use another namespace, replace `kubeflow-user-example-com` everywhere in
+`manifests.yaml` — including inside `KUBEFLOW_MCP_ALLOWED_HOSTS`, which spells out
+the Service DNS names:
+
+```bash
+sed -i 's/kubeflow-user-example-com/my-namespace/g' manifests.yaml
+```
+
+## Choosing what the agent can do
+
+Tool exposure is controlled by the persona, which is separate from RBAC — RBAC
+decides what the ServiceAccount may touch in the Kubernetes API, the persona
+decides which MCP tools exist at all.
+
+| `KUBEFLOW_MCP_PERSONA` | Tools |
+|---|---|
+| `readonly` (default) | Inspection and monitoring only — cannot submit a TrainJob |
+| `data-scientist` (used here) | `readonly` plus `fine_tune`, `run_custom_training`, `wait_for_training`, `delete_training_job` |
+| `ml-engineer` | `data-scientist` plus `run_container_training`, `update_training_job`, `inspect_crd`, `inspect_controller` |
+| `platform-admin` | Everything |
+
+The example uses `data-scientist` because the default `readonly` persona cannot
+submit training jobs, which is the first thing most people want to try. Switch the
+`KUBEFLOW_MCP_PERSONA` env var to change it.
+
+`ml-engineer` additionally reads the Trainer controller's pods and logs, which the
+RBAC in this example does not grant. If you use it, add a read-only Role for the
+controller namespace and set `KUBEFLOW_MCP_CONTROLLER_NAMESPACE`.
+
+For finer control than the built-in personas, mount a policy file at
+`$HOME/.kf-mcp-policy.yaml` in the pod — see the main
+[README](../../README.md) for the format.
+
+## Troubleshooting
+
+**Requests return HTTP 421** — DNS rebinding protection allows loopback `Host`
+headers by default, so anything arriving via the Service is rejected.
+`KUBEFLOW_MCP_ALLOWED_HOSTS` must list the exact hostname clients use; only the
+`:*` port wildcard is supported, not a host wildcard. Note that `port-forward`
+does *not* reproduce this, because it sends a loopback `Host` header.
+
+**Tools report an empty namespace, or a 403 listing runtimes** — the pod is running
+outside the namespace whose TrainJobs you expect. See the section above.
+
+**The agent has no tool for submitting a TrainJob** — the persona is `readonly`.
+Set `KUBEFLOW_MCP_PERSONA` to `data-scientist` or higher.
+
+**Log line: `Trainer control-plane version info is not available ... (404)`** — the
+SDK reads the Trainer version from the `kubeflow-trainer-public` ConfigMap in
+`kubeflow-system`, which older Trainer releases do not create. Harmless: every
+`check_compatibility` check still passes, including the CRD and API version. If the
+same message shows `(403)` instead, the `kubeflow-mcp-trainer-version` Role is
+missing or Trainer runs in a different namespace than `kubeflow-system`.
+
+**Every authenticated request returns 401** — the token still reads `REPLACE_ME`,
+or the client is sending a different one. Compare against the value stored in the
+Secret.
+
+**Pod stuck in `CreateContainerConfigError`** — the `kubeflow-mcp-auth` Secret is
+missing, which happens if you removed it from `manifests.yaml` without creating it
+imperatively.
--- a/examples/kubernetes/README.md
+++ b/examples/kubernetes/README.md
@@ -7,22 +7,20 @@ that agents can reach it over HTTP and drive
 It is intentionally small: one Deployment, one ClusterIP Service, a ServiceAccount
 and least-privilege RBAC. No Ingress, no OIDC, no Helm chart.
 
-## Prerequisites
-
-- A Kubernetes cluster with Kubeflow Trainer installed
-- A namespace to run in. The example uses `kubeflow-user-example-com`, the Profile
-  namespace created by a standard Kubeflow installation
-- `kubectl` pointed at that cluster
-
 ## Quick start
+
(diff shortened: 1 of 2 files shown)


Krishna-kg732:
Hey @szymka117 , thanks for opening this 😃 , will review it shortly

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kubeflow-mcp-auth
+  namespace: kubeflow-user-example-com
+  labels:
+    app.kubernetes.io/name: kubeflow-mcp
+type: Opaque
+stringData:
+  token: REPLACE_ME
```
Krishna-kg732: Shipping a functional default token makes the obvious command the unsafe one. kubectl apply -f manifests.yaml — which people will try before reading the quick start — brings up a server whose bearer token is published in this repo, with the data-scientist persona (create/delete TrainJobs), a cluster-reachable ClusterIP, and sidecar.istio.io/inject: "false" so no Istio AuthorizationPolicy backstops it. Any pod in the cluster can then drive it.

Rather than documenting around this, consider dropping the Secret document entirely and making imperative creation a required step. A missing Secret gives CreateContainerConfigError — loud and safe — instead of a quietly-open server. It'd also remove the "leaving the document in place resets the token" caveat at README:45 and the token-churn problem below.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+    resources: ["nodes", "namespaces"]
+    verbs: ["get", "list"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["get", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kubeflow-mcp-read
```
Krishna-kg732: ClusterRole and ClusterRoleBinding are cluster-scoped but named with a fixed string, while README:106's sed only rewrites the namespace. So a second install silently breaks the first: team B applies, ClusterRoleBinding/kubeflow-mcp-read is overwritten, and subjects now contains only team B's ServiceAccount. Team A's server starts 403ing on list_runtimes/get_cluster_resources with nothing changed on their side.

Suggest suffixing the cluster-scoped names with the namespace (kubeflow-mcp-read-kubeflow-user-example-com) so the documented sed renames them too.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+    resources: ["trainingruntimes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["jobset.x-k8s.io"]
+    resources: ["jobsets"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["batch"]
+    resources: ["jobs"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["pods", "pods/log", "events", "persistentvolumeclaims"]
```
Krishna-kg732, line 57: get_runtime(include_packages=True) creates and deletes a Pod ([discovery.py:313](https://github.com/kubeflow/mcp-server/blob/main/kubeflow_mcp/trainer/api/discovery.py#L313) and :349), but this Role grants pods only get/list/watch, so it always 403s. That tool is in the readonly persona, so it's exposed in every configuration this example supports.

```suggestion
  - apiGroups: [""]
    resources: ["pods", "pods/log", "events", "persistentvolumeclaims"]
    verbs: ["get", "list", "watch"]
  # get_runtime(include_packages=True) creates and deletes a probe Pod
  - apiGroups: [""]
    resources: ["pods"]
    verbs: ["create", "delete"]
```

Alternatively, note in the README that include_packages=True isn't supported by this example's RBAC.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: kubeflow-mcp
+    namespace: kubeflow-user-example-com
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: kubeflow-mcp-trainer-version
+  namespace: kubeflow-system
```
Krishna-kg732, line 81: These two documents hardcode kubeflow-system, which the manifest never creates and README:106's sed never rewrites. If Trainer is installed elsewhere, kubectl apply -f - partially fails with namespaces "kubeflow-system" not found — Deployment and Service get created, the version-read RBAC doesn't. The troubleshooting section covers the resulting runtime 403 but not this apply-time failure; worth a line there.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+          image: ghcr.io/kubeflow/mcp-server:latest
+          ports:
+            - name: http
+              containerPort: 8000
+          env:
+            - name: MCP_TRANSPORT
+              value: http
+            - name: FASTMCP_HOST
+              value: "0.0.0.0"
+            - name: KUBEFLOW_MCP_ALLOWED_HOSTS
```
Krishna-kg732, line 138: Only ALLOWED_HOSTS is set. allowed_origins then falls back to loopback-only defaults ([transport_security.py:81](https://github.com/kubeflow/mcp-server/blob/main/kubeflow_mcp/core/transport_security.py#L81), defaults at 62-69), so a browser-based client at http://kubeflow-mcp:8000 sends a matching Origin and gets 403 — a different code and cause from the 421 the README explains. Non-browser clients send no Origin and are unaffected, which makes this easy to miss. Either set KUBEFLOW_MCP_ALLOWED_ORIGINS alongside the hosts list, or add the 403 to troubleshooting.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+              value: "kubeflow-mcp:*,kubeflow-mcp.kubeflow-user-example-com.svc:*,kubeflow-mcp.kubeflow-user-example-com.svc.cluster.local:*,localhost:*,127.0.0.1:*"
+            - name: KUBEFLOW_MCP_PERSONA
+              value: data-scientist
+            - name: KUBEFLOW_MCP_AUTH_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: kubeflow-mcp-auth
+                  key: token
+          livenessProbe:
```
Krishna-kg732, line 159:
```suggestion
          startupProbe:
            httpGet:
              path: /health
              port: http
            failureThreshold: 30
            periodSeconds: 2
          livenessProbe:
            httpGet:
              path: /health
              port: http
          readinessProbe:
            httpGet:
              path: /ready
              port: http
```

Krishna-kg732: Probes rely entirely on defaults: initialDelaySeconds: 0, timeoutSeconds: 1, periodSeconds: 10, failureThreshold: 3 — about 30s to first serve, against a 50m CPU request. Server construction is ~1s on an idle machine, so this is fine normally; under node contention the guaranteed share is 1/20th of a core and startup can overrun into CrashLoopBackOff. It won't reproduce in a kind cluster, which is what makes it worth guarding:

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+            httpGet:
+              path: /health
+              port: http
+          readinessProbe:
```
Krishna-kg732, line 166: Minor: /ready binds is_ready at route-registration time ([http_edge.py:22-33](https://github.com/kubeflow/mcp-server/blob/main/kubeflow_mcp/core/http_edge.py#L22-L33)) and never re-evaluates, so it returns 200 for the life of the process and carries no more signal than /health. If the Kubernetes API becomes unreachable the pod stays in the Service endpoints. Not this PR's bug, but the manifest presents it as a real dependency check — maybe worth a README note.

examples/kubernetes/README.md
Krishna-kg732, line 1: "no credential is ever written to disk" isn't quite right. kubectl apply is a client-side apply, so it writes the submitted manifest into kubectl.kubernetes.io/last-applied-configuration on the Secret, containing "stringData":{"token":"<real token>"} in cleartext. Unlike the base64 data field, that annotation is echoed verbatim by kubectl get secret -o yaml and captured in backups and GitOps diffs. Suggest narrowing to "keeps the token out of your local manifests.yaml".

examples/kubernetes/README.md
```diff
@@ -0,0 +1,161 @@
...
+kubectl rollout status deploy/kubeflow-mcp -n "$NAMESPACE"
+```
+
+Read the token back when configuring a client:
+
+```bash
+kubectl get secret kubeflow-mcp-auth -n "$NAMESPACE" -o jsonpath='{.data.token}' | base64 -d
+```
+
+To rotate it later, re-run the command above and restart the pod so the new value
```
Krishna-kg732: Worth calling out that the pipeline regenerates the token on every run, not just intentional rotations. Someone editing KUBEFLOW_MCP_PERSONA and re-running the quick-start command rotates the credential as a side effect and 401s every configured client after the next restart. The reset warning at line 45 only covers the removed-Secret case.

examples/kubernetes/README.md
```diff
@@ -0,0 +1,161 @@
...
+Kubeflow SDK derives its default namespace from the pod's ServiceAccount namespace.
+Running it elsewhere makes every tool default to a namespace that has no TrainJobs
+in it.
+
+To use another namespace, replace `kubeflow-user-example-com` everywhere in
+`manifests.yaml` — including inside `KUBEFLOW_MCP_ALLOWED_HOSTS`, which spells out
+the Service DNS names:
+
+```bash
+sed -i 's/kubeflow-user-example-com/my-namespace/g' manifests.yaml
```
Krishna-kg732: sed -i 's/.../.../g' file is GNU-only; on macOS, BSD sed reads s/... as the backup suffix and fails with sed: 1: "manifests.yaml": undefined label. Since kubectl workstations are frequently Macs: sed -i.bak 's/kubeflow-user-example-com/my-namespace/g' manifests.yaml.

Krishna-kg732 (changes requested):
Nice example — the manifests are clean and the README is more thorough than most. I verified the DNS-rebinding explanation against the source and it's accurate: /health and /ready really are exempt from Host validation ([transport_security.py:95](https://github.com/kubeflow/mcp-server/blob/main/kubeflow_mcp/core/transport_security.py#L95)), so the probe story holds up.

Most of my comments are in the gap between what the README promises and what the code does. Three I'd want addressed before merge: the shipped default token, the cluster-scoped RBAC name collision, and get_runtime(include_packages=True) being unusable under this Role. The rest are smaller.

One scoping note for other reviewers: I checked whether the absent Katib RBAC mattered, and it doesn't — create_server defaults to clients=["trainer"] ([config.py:87](https://github.com/kubeflow/mcp-server/blob/main/kubeflow_mcp/core/config.py#L87)), so the optimizer tools never register. The persona table is accurate as written.

(Reviewed by Claude Opus 5 PTAL)

examples/kubernetes/manifests.yaml
HarshPopat23, line 1: The Role currently grants `create`, `update`, `patch`, and `delete` on `trainjobs/status` because it is grouped with `trainjobs`. The status subresource is controller-owned, and the MCP server should only mutate the TrainJob specification. Normal TrainJob reads already include status, so could we remove `trainjobs/status` from this rule? That would better match the PR’s least-privilege goal and prevent the ServiceAccount from modifying controller-reported job state.

HarshPopat23, line 1: Since `kubeflow-user-example-com` is described as an existing Profile namespace, should the Namespace document be removed from this file? Profile namespaces are reconciled by the Profile controller, and including the Namespace here means a later `kubectl delete -f manifests.yaml` could delete the user’s entire Profile namespace and unrelated workloads. I think it would be safer to make the namespace a prerequisite, with a separate `kubectl create namespace` step only for standalone local Trainer clusters.

HarshPopat23 (commented):
Other then two inline comment i think below is another thing need to improve.
One validation detail: the PR says the example was tested with Trainer v2.1.0, while the repository compatibility table currently requires Trainer >=2.2.0. Since `pre_flight` mainly validates Kubernetes and the Trainer CRD/API version, its passing result does not necessarily establish support for the older release. Could the example be rerun with a supported Trainer version, or could the PR description clarify that v2.1.0 was only the environment used for incidental testing?

szymka117:
Hi, let me improve the PR. I'll bump the version and improve resource definition then return the tests.

abhijeet-dhumal:
Thanks @szymka117 this is exactly what we need. In-cluster deployment is one of the most asked-about gaps.
Please address the blocking comments above and we'll get this merged. No rush!
Adding some additional nits 👀

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+      serviceAccountName: kubeflow-mcp
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65532
+        runAsGroup: 65532
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: server
+          image: ghcr.io/kubeflow/mcp-server:latest
```
abhijeet-dhumal: Can we pin to a release tag for reproducibility. latest in an example that people copy-paste might  silently break when the image updates.
WDYT?

examples/kubernetes/manifests.yaml
abhijeet-dhumal: Agreed, @szymka117  consider 100m minimum. At 50m under contention, python startup + uv sync can exceed the startup probe window even with the suggested startupProbe fix.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+              port: http
+          resources:
+            requests:
+              cpu: 50m
+              memory: 128Mi
+            limits:
+              memory: 512Mi
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
```
abhijeet-dhumal, line 181: I think the Kubeflow SDK and HuggingFace Hub client write to ~/.cache/huggingface. With runAsUser: 65532 + readOnlyRootFilesystem: true, any tool that touches the HF cache will crash.

abhijeet-dhumal: Lets try adding either an emptyDir at /home/nonroot/.cache or document that HF-dependent tools (fine_tune with gated models) won't work without it.

examples/kubernetes/manifests.yaml
```diff
@@ -0,0 +1,199 @@
...
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kubeflow-mcp
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: kubeflow-mcp
+      annotations:
+        sidecar.istio.io/inject: "false"
```
abhijeet-dhumal, line 118: this setting is fine for the example, but worth a doc line: "If your mesh enforces strict mTLS, remove this annotation and configure a PeerAuthentication/DestinationRule instead."

abhijeet-dhumal:
Nit: It seems the README is thorough but oversized for an example.
Consider trimming to only sections : quick start, resource table, verify, namespace change, general troubleshooting.

abhijeet-dhumal:
Hey @szymka117 are you still working on this ?

szymka117:
Hey, yes almost done. I am running some last tests but I will have an update today.

examples/kubernetes/manifests.yaml
szymka117: Fixed

szymka117: Fixed

examples/kubernetes/manifests.yaml
szymka117: Fixed

examples/kubernetes/manifests.yaml
szymka117: Yes, it was designed in mind to be used with kubeflow platform provided trainer therefor kubeflow-system. added a comment in troubleshooting section

examples/kubernetes/manifests.yaml
szymka117: Note added in readme

examples/kubernetes/manifests.yaml
szymka117: Added.

examples/kubernetes/manifests.yaml
szymka117: Increased req.

examples/kubernetes/manifests.yaml
szymka117: Added.

examples/kubernetes/manifests.yaml
szymka117: Added.

examples/kubernetes/README.md
szymka117: Fixed, it was rewritten on occasion of token secret handling comment.

szymka117:
Added new version fixing i think majority of the comments.
I've tried to get rid of the sed in favour of kustomize but there didn't seem to be an easy way to do it and I didn't want to over generate and complicate code.
Also i have created small dev helper for testing but didn't include it in the PR. It might be helpful for testing:
https://github.com/szymka117/mcp-server/tree/feat/k8s-deployment-manifests-e2e-tests/hack
It was tested only on Linux but can deploy full kind cluster with kubeflow and run manifest with some basic tests against it.

abhijeet-dhumal:
Thanks @szymka117, this is great contribution ! 🚀 
/lgtm
/approve

PR #167, opened 2026-08-24
fix(trainer): bound memory when reading training job logs

### LINKED ISSUE #166: fix(trainer): bound log line memory consumption in get_training_logs
### Problem Description

In `kubeflow_mcp.trainer.api.monitoring.get_training_logs()`, log lines from `TrainerClient.get_job_logs()` are consumed into memory with `list()`:

```python
client = get_trainer_client_for_namespace(namespace)
log_lines = list(client.get_job_logs(name=name, step=step, follow=False))
```

`TrainerClient.get_job_logs()` returns an `Iterator[str]`. For long-running training jobs that emit a large volume of logs, calling `list()` forces the entire generator to be evaluated into an unbounded Python list in memory before the subsequent slice (`log_lines = log_lines[-MAX_LOG_LINES:]`) can truncate it.

### Expected Behavior

Log lines should be consumed using a bounded sliding window (`collections.deque(..., maxlen=MAX_LOG_LINES)`) so that memory usage in the MCP server remains bounded to `MAX_LOG_LINES` throughout iteration.

### Proposed Solution

```python
from collections import deque

log_lines = list(
    deque(client.get_job_logs(name=name, step=step, follow=False), maxlen=MAX_LOG_LINES)
)
```

priyank766:
### What this PR does

- Consumes `client.get_job_logs()` using `collections.deque(..., maxlen=MAX_LOG_LINES)` instead of an unbounded `list()`.
- Keeps log memory in the MCP server bounded to `MAX_LOG_LINES` during stream iteration, avoiding large intermediate list allocations on high-output training jobs.
- Adds `test_log_truncation_with_generator` in `kubeflow_mcp/trainer/api/monitoring_test.py` with boundary line count assertions.

Fixes #166

### How it was tested

- `uv run pytest kubeflow_mcp/trainer/api/monitoring_test.py` (45 passed)
- `uv run pytest` (480 passed)
- `make verify` (clean)

### DIFF
--- a/kubeflow_mcp/trainer/api/monitoring.py
+++ b/kubeflow_mcp/trainer/api/monitoring.py
@@ -16,6 +16,7 @@
 
 import logging
 import re
+from collections import deque
 from typing import Any
 
 from kubeflow_mcp.common.constants import ErrorCode
@@ -168,7 +169,9 @@ def get_training_logs(
             ).model_dump()
 
         client = get_trainer_client_for_namespace(namespace)
-        log_lines = list(client.get_job_logs(name=name, step=step, follow=False))
+        log_lines = list(
+            deque(client.get_job_logs(name=name, step=step, follow=False), maxlen=MAX_LOG_LINES)
+        )
         if not log_lines:
             try:
                 eff_ns = get_trainer_effective_namespace(namespace)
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
Thanks @priyank766, great catch! 
deque with maxlen for the generator is the right fix 🙌 
/lgtm
/approve

PR #170, opened 2026-08-25
feat(test): add MCP tool schema snapshot test

Kartikeya-trivedi:
## What this PR does

Adds a snapshot test for the MCP `tools/list` surface, per #168. The test captures tool names, parameters, types, required/optional fields, output schemas, and behavior annotations for the `platform-admin` persona (which exposes all tools) into a committed JSON baseline, and fails with a unified diff if the surface drifts without an explicit snapshot update.

This turns the AGENTS.md schema stability rule (no tool schema changes without KEP / maintainer approval) into a CI gate — analogous to OpenAPI spec diffing for REST APIs.

## How it works

- `tests/conformance/tool_schema_snapshot_test.py` builds the server via `create_server(persona="platform-admin")` and lists tools through an **in-memory FastMCP client** — no cluster, no network, no subprocess.
- `tests/conformance/snapshots/tool_schema_platform_admin.json` is the committed baseline (23 tools).
- The snapshot is **structural only**: tool descriptions, titles, and per-parameter `description` text are stripped, so documentation wording edits do not churn the baseline. Only changes to names, parameters, types, required/optional fields, output schemas, or annotations trip the gate.
- `make update-schema-snapshot` regenerates the baseline for intentional, approved changes.

## Example failure output

```
Failed: MCP tool schema surface changed.

--- baseline (committed)
+++ current (tools/list)
@@ -1,5 +1,5 @@
 {
-  "check_compatibility": {
+  "check_compat_RENAMED": {
...
Tool schemas are a stability contract (see AGENTS.md): changes to tool
names, parameters, types, or required fields need KEP / maintainer
approval. If this change is intentional and approved, refresh the
baseline with:

    make update-schema-snapshot
```

## Acceptance criteria (from #168)

- [x] Runs in `make test-python` — no cluster, no network, fast (in-memory client; full suite 499 passed in ~12s)
- [x] Clear failure message with diff on schema change
- [x] `make update-schema-snapshot` for intentional updates
- [x] Covers `platform-admin` persona (exposes all tools)

Fixes #168

PR #147, opened 2026-08-05
fix(docs): update SECURITY.md dependency scanning to OSV-Scanner

Kartikeya-trivedi:
## Description

One-line docs fix: SECURITY.md (merged via #14) still says `pip-audit` runs in CI, but pip-audit was replaced by the nightly OSV-Scanner workflow in #30. This updates the Prevention Mechanisms line to describe the current setup — nightly scan of `uv.lock`, findings in the Security tab, automated fix PRs (e.g. #137).

Follow-up promised on the #30 review thread.

## Type of Change

- [ ] feat: New feature
- [x] fix: Bug fix
- [ ] revert: Revert a change
- [ ] chore: Maintenance / tooling

## Checklist

- [x] Linting passes (`make verify`)
- [x] Documentation updated (if applicable)
- [x] Commit messages follow conventional format

## Related Issues

Follow-up to #30 / #14

PR #172, opened 2026-08-25
fix(ci): stop welcoming returning contributors as first-timers

### LINKED ISSUE #171: bug(ci): welcome workflow greets returning contributors as first-timers on every PR
## What happened

The `Welcome new contributors` workflow posted the *"Thanks for opening your first PR!"* message on #90, #147, and #170 — all opened by the same returning contributor (me), whose first PR #30 was merged back in July. The bot greets me as a first-timer on every PR I open.

## Root cause

The workflow uses `actions/first-interaction@v3`, whose deployed logic (`dist/index.js` at the `v3` tag) has two stacked bugs:

1. **`isFirstPullRequest()` never filters by author.** It lists *all* PRs in the repo and returns true only if no PR with a lower number exists — i.e. it can only be true for the repo's very first PR, regardless of who opened it:

   ```js
   const pulls = await octokit.paginate(octokit.rest.pulls.list, {
     owner: ..., repo: ..., state: 'all'   // no creator/author filter
   });
   return pulls.filter((pull) => pull.number < context.issue.number).length === 0;
   ```

2. **The skip guard requires BOTH checks to say "not first":**

   ```js
   if (!(await isFirstIssue(octokit)) && !(await isFirstPullRequest(octokit)))
     return core.info('Skipping...Not First Contribution');
   ```

   On a **PR** event it still consults `isFirstIssue`, which *does* filter by author. So any contributor who has opened PRs but never filed an issue in this repo makes `isFirstIssue` return true, the skip is bypassed, and the `pr_message` is posted — on **every** PR they open. That's exactly my case: 4 PRs, 0 issues.

This matches upstream report actions/first-interaction#369 ("Every contributor identified as new contributor").

## Proposed fix

Since the bug is in the action itself (still broken at `v3`/`main`), replace `actions/first-interaction` with a small `actions/github-script` step that checks the actual author's history via the search API (`repo:… type:pr author:…` / `type:issue`), keeps the same welcome messages, skips bots, and drops the unnecessary `contents: write` permission. PR incoming.

/area testing

Kartikeya-trivedi:
## What this PR does

Fixes the `Welcome new contributors` workflow greeting returning contributors with *"Thanks for opening your first PR!"* on **every** PR they open (happened on #90, #147, and #170 — all mine, well after #30 was merged).

## Root cause

`actions/first-interaction@v3` is broken (see #171 for the full analysis, and upstream actions/first-interaction#369):

1. Its `isFirstPullRequest()` lists **all** repo PRs with no author filter, so it can only be true for the repo's literal first PR.
2. Its skip guard is `if (!isFirstIssue && !isFirstPullRequest) skip` — on a PR event it still consults `isFirstIssue`, which **is** author-filtered. Any PR author who has never filed an issue in this repo makes that check true, bypassing the skip, so the `pr_message` posts on every PR they open.

## The fix

Replace `actions/first-interaction` with a small `actions/github-script` step that:

- searches the **author's own history** in this repo (`type:pr author:X` / `type:issue author:X`) and only comments when no earlier item by them exists
- keeps the exact same welcome messages (moved to `env` so no script injection surface)
- skips bot senders (e.g. google-oss-prow, dependabot), which the old action greeted too
- drops the unneeded `contents: write` permission (least privilege — the job only comments)

The workflow keeps running on `pull_request_target` so it can comment on fork PRs; the script never checks out or executes PR code, so that trigger remains safe.

Fixes #171

### DIFF
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -9,7 +9,6 @@ on:
       - opened
 
 permissions:
-  contents: write
   issues: write
   pull-requests: write
 
@@ -17,10 +16,14 @@ jobs:
   welcome:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/first-interaction@v3
-        with:
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          issue_message: |
+      # actions/first-interaction@v3 greets returning contributors on every
+      # PR (its PR check ignores the author, and its skip guard lets any PR
+      # author who has never filed an issue through — see issue #171 and
+      # https://github.com/actions/first-interaction/issues/369). Check the
+      # author's own history via the search API instead.
+      - uses: actions/github-script@v8
+        env:
+          ISSUE_MESSAGE: |
             🎉 **Welcome to the Kubeflow MCP Server!** 🎉
 
             Thanks for opening your first issue! We're happy to have you as part of our community 🚀
@@ -37,7 +40,7 @@ jobs:
             Feel free to ask questions in the comments if you need any help or clarification!
             Thanks again for contributing to Kubeflow! 🙏
 
-          pr_message: |
+          PR_MESSAGE: |
             🎉 **Welcome to the Kubeflow MCP Server!** 🎉
 
             Thanks for opening your first PR! We're happy to have you as part of our community 🚀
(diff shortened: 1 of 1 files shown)


abhijeet-dhumal:
Thanks @Kartikeya-trivedi !
/lgtm
/approve

PR #189, opened 2026-08-25
fix(ci): pin Trainer ref in E2E workflow to mitigate cache poisoning

### LINKED ISSUE #188: fix(ci): pin Trainer ref in E2E workflow to mitigate cache poisoning (CodeQL alert #10)
### Problem

CodeQL [alert #10](https://github.com/kubeflow/mcp-server/security/code-scanning/10) flags `.github/workflows/e2e-k8s.yaml` for potential cache poisoning:

- The workflow checks out `kubeflow/trainer` at `master` (mutable ref)
- Runs `make test-e2e-setup-cluster` from that checkout
- uv cache is enabled and persists on the default branch context via `workflow_dispatch`

A compromised commit on `kubeflow/trainer:master` could inject malicious code into cached dependencies.

### Severity

High (CodeQL) / Low practical risk (same-org sibling repo, same maintainers).

### Fix

1. Pin `trainer-ref` to a release tag instead of `master`
2. Optionally disable uv cache for the E2E job (`enable-cache: false`) since it runs infrequently
3. Add `persist-credentials: false` to the trainer checkout step

```yaml
matrix:
  kubernetes-version: ["1.32.3"]
  trainer-ref: ["v2.2.0"]
```

### Acceptance criteria

- [ ] trainer-ref pinned to a stable release tag
- [ ] CodeQL alert #10 resolves after merge
- [ ] E2E tests still p ass with the pinned ref

/kind bug
/area security
/area engprod
/good-first-issue

myselfkunal:
## Description

Pins the Kubeflow Trainer checkout in the E2E workflow to a stable release tag
instead of the mutable `master` ref, addressing the cache-poisoning risk flagged
by CodeQL alert #10 on `.github/workflows/e2e-k8s.yaml`.

Three changes, all in `.github/workflows/e2e-k8s.yaml`:

1. `trainer-ref` pinned from `master` to `v2.3.0`
2. `enable-cache: false` on the `astral-sh/setup-uv` step for the E2E job
3. `persist-credentials: false` added to the Trainer checkout step

Two notes for reviewers:

- The issue suggests `v2.2.0`, but I went with `v2.3.0` , it's the latest stable
  Trainer release (Aug 2026) and the closest to current `master`, so it's the
  lowest-risk pin for keeping E2E green. `test-e2e-setup-cluster` exists in the
  `v2.3.0` Makefile. Happy to switch to `v2.2.0`/`v2.2.1` if you'd prefer.
- Included `enable-cache: false` (listed as optional in the issue) because
  CodeQL's cache-poisoning rule flags the cache write itself, so pinning the ref
  alone may not resolve alert #10. The E2E job runs infrequently, so the cache
  buys little here.

## Related Issue

Fixes #188

## Checklist

## Testing

CI-only change; no application code touched.

- [ ] Unit tests
- [ ] Integration tests
- [x] E2E tests

Verified `test-e2e-setup-cluster` is present in the `kubeflow/trainer` Makefile
at `v2.3.0` and that the YAML changes are limited to the three lines above.

The E2E job is gated behind the `ok-to-test-e2e` label so could a maintainer add
it so the pinned ref gets a real run? That covers the third acceptance criterion.

### DIFF
--- a/.github/workflows/e2e-k8s.yaml
+++ b/.github/workflows/e2e-k8s.yaml
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         kubernetes-version: ["1.32.3"]
-        trainer-ref: ["v2.2.0"]
+        trainer-ref: ["v2.2.1"]
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/e2e-k8s.yaml
+++ b/.github/workflows/e2e-k8s.yaml
@@ -36,11 +36,13 @@ jobs:
       fail-fast: false
       matrix:
         kubernetes-version: ["1.32.3"]
-        trainer-ref: ["master"]
+        trainer-ref: ["v2.2.0"]
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
+        with:
+          persist-credentials: false
 
       - name: Checkout Kubeflow Trainer repository
         uses: actions/checkout@v7
@@ -48,6 +50,7 @@ jobs:
           repository: kubeflow/trainer
           ref: ${{ matrix.trainer-ref }}
           path: trainer
+          persist-credentials: false
 
       - name: Set up Python
         uses: actions/setup-python@v7
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ claude mcp add kubeflow -- kubeflow-mcp serve
 
 | MCP Server | Kubeflow Trainer | Kubeflow SDK | Python      | Kubernetes |
 |------------|------------------|--------------|-------------|------------|
-| 0.1.x      | >= 2.2.0         | >= 0.4.0     | 3.10 - 3.12 | >= 1.27    |
+| 0.1.x      | == 2.2.0         | >= 0.4.0     | 3.10 - 3.12 | >= 1.27    |
 
 ## CLI Reference
 
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ claude mcp add kubeflow -- kubeflow-mcp serve
 
 | MCP Server | Kubeflow Trainer | Kubeflow SDK | Python      | Kubernetes |
 |------------|------------------|--------------|-------------|------------|
-| 0.1.x      | == 2.2.0         | >= 0.4.0     | 3.10 - 3.12 | >= 1.27    |
+| 0.1.x      | == 2.2.1         | >= 0.4.0     | 3.10 - 3.12 | >= 1.27    |
 
 ## CLI Reference
 


jaiakash (commented):
For the security warning, pinning to good (`2.3.0` version should be good).

Can we confirm why we need to disable the `cache` and `persist-credentials`? 
- `cache` is perfectly fine and may help even if we are running the action infrequently.
- `persist-credentials` — the job is `contents: read` on a public repo, so i dont see any issue. But if we are removing it, lets remove from the first checkout as well?

wdyt ?

myselfkunal:
Thanks for the review! @jaiakash 

**cache** - fair point, happy to revert this. For context on why I included it:
this workflow runs on `push: main` and `workflow_dispatch`, so the uv cache is
written into the default-branch scope, and any PR run can restore from that
scope. The cache save happens post-job, i.e. after `make test-e2e-setup-cluster`
has executed code from the trainer checkout, so with `master` there was a path
from "bad commit in trainer" to "poisoned cache restored by every PR". But
pinning the ref removes the untrusted source, which is the real fix, so the cache
disable is belt-and-braces. Reverting to `enable-cache: true`.

One caveat worth flagging: git tags are mutable, so `v2.3.0` could in principle be
re-pointed. If you want the stronger guarantee we could pin the commit SHA with
the tag in a trailing comment instead, probably overkill for a same-org sibling
repo, but happy to if you'd prefer.

**persist-credentials** - agreed the practical risk is negligible with
`contents: read` on a public repo. I included it because it was item 3 in #188.
Mild preference for your suggestion of adding it to both checkouts rather than
dropping it: the trainer checkout is the one whose code we execute, so not
leaving the token in `trainer/.git/config` is cheap defense-in-depth, and doing
both is more consistent than doing neither. But happy to go either way, let me
know what you'd prefer and I'll push it.

Proposed update: revert the cache change, add `persist-credentials: false` to
both checkout steps, keep the `v2.3.0` pin. Then we can see whether alert #10
actually clears on merge, if not I'll follow up.

myselfkunal:
@jaiakash @abhijeet-dhumal 
Thanks for the review! I've pushed the update: reverted the cache change back
to `enable-cache: true`, and added `persist-credentials: false` to both
checkout steps. Let me know if you'd rather handle it differently, happy to
adjust further.

abhijeet-dhumal:
Hey @myselfkunal , can we stick to Trainer v2.2.0 for now .. 
There are some breaking changes introoduced in later version for which MCP server is not compatible.. 
and also make sure to update readme to point existing Trainer's compatibility metrics to ==v2.2.0 instead of >=v2.2.0

myselfkunal:
hey @abhijeet-dhumal !
thanks for the heads up. Pushed the revert: `trainer-ref` back to
`v2.2.0`, and updated the README compatibility table to `== 2.2.0` instead of
`>= 2.2.0`.

abhijeet-dhumal:
Thanks @myselfkunal, one last thing : to validate this PR we need to enable E2E test pr check, 
E2E is SKIPPED because ok-to-test-e2e label addition isn't practical.. Prow bot won't apply it via /label.
Can you drop the` job if: label check` in e2e-k8s.yaml? Path filter is enough. Then this PR should run E2E on sync.

I validated locally: Trainer v2.2.0 + make test-e2e passes on Kind.
Just  need to remove this if block : 
```
 if: >-
      github.event_name != 'pull_request' ||
      contains(github.event.pull_request.labels.*.name, 'ok-to-test-e2e')
```
can you please do this ? 
and once E2E PR checks passes.. we can merge this PR?

abhijeet-dhumal:
@myselfkunal Apologies I had to close and reopen your PR to run e2e test pr check !
Thanks!
/lgtm
/approve

myselfkunal:
> /retest

Looked into the Setup cluster failure, it's not related to this PR's changes.

The E2E job fails at `hack/e2e-setup-cluster.sh` with:

    jobset-controller-manager-8667ff49bd-tg5w7   0/1   ImagePullBackOff
    Image: us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset:v0.11.0

This is JobSet's staging image registry, not registry.k8s.io, and it's not meant for long-term pulls. Looks like it's been GC'd or is currently inaccessible. trainer-controller-manager builds and starts fine, only the jobset sidecar fails to pull.

Current upstream JobSet release is v0.12.0, and its manifests still use the same staging path pattern, so this will probably keep failing on retest until the pinned jobset version/image source in the e2e manifests gets updated.

abhijeet-dhumal:
Hey @myselfkunal, can you rebase this PR? 
I have update oracle runner's to CNCF which will unblock E2E test PR check run : https://github.com/kubeflow/mcp-server/pull/199

.github/workflows/e2e-k8s.yaml
```diff
@@ -36,18 +41,21 @@ jobs:
       fail-fast: false
       matrix:
         kubernetes-version: ["1.32.3"]
-        trainer-ref: ["master"]
+        trainer-ref: ["v2.2.0"]
```
abhijeet-dhumal:
```suggestion
        trainer-ref: ["v2.2.1"]
```
Hey @myselfkunal , Here we need to update this version to patch release.. this will resolve E2E test PR check !

Krishna-kg732:
Just bump the version and this should be good to go 
rest lgtm ,Thanks for picking this up

CC: @myselfkunal @abhijeet-dhumal

Krishna-kg732:
@myselfkunal , can you please dco sign your commit (use git commit -s --amend --no-edit)

myselfkunal:
> @myselfkunal , can you please dco sign your commit (use git commit -s --amend --no-edit)

Apologies for that, missed the sign-off in the rush. Fixed it now, thanks for flagging @Krishna-kg732 !

abhijeet-dhumal:
Thanks @myselfkunal !
/hold cancel
/approve

PR #192, opened 2026-08-31
fix(trainer): extract runtime metadata from SDK Runtime in get_runtime

### LINKED ISSUE #191: fix(trainer): extract runtime metadata from SDK Runtime object in get_runtime
### Problem Description

In `kubeflow_mcp.trainer.api.discovery.get_runtime()`, metadata extraction attempts to read runtime specifications from `rt.spec`:

```python
client = get_trainer_client()
rt = client.get_runtime(name=name)

spec = getattr(rt, "spec", None)
if spec is not None:
    ml_policy = getattr(spec, "ml_policy", None)
    data["framework"] = getattr(ml_policy, "torch", None) or getattr(ml_policy, "mpi", None)
    ...
```

However, `TrainerClient.get_runtime()` from the Kubeflow SDK returns a `kubeflow.trainer.types.types.Runtime` dataclass with `rt.name`, `rt.trainer` (`RuntimeTrainer`), and `rt.pretrained_model`. The SDK object does not have a `.spec` attribute.

Because `getattr(rt, "spec", None)` is `None`, `get_runtime()` drops `framework`, `image`, `device`, `trainer_type`, and `num_nodes` on every invocation against a cluster, returning only `{"name": rt_name}`.

### Expected Behavior

`get_runtime()` should extract metadata from `rt.trainer` (`framework`, `image`, `num_nodes`, `device`, `device_count`, `trainer_type`) and `rt.pretrained_model`, while preserving the `rt.spec` fallback for raw CRD / mock objects.

### Proposed Solution

1. Extract runtime details from `rt.trainer` and `rt.pretrained_model`.
2. Update `_get_runtime_image` to check `client.get_runtime(name)` first (which resolves both namespaced and cluster-scoped runtimes) before falling back to `CustomObjectsApi`.

priyank766:
### What this PR does

- Updates `get_runtime()` to extract runtime metadata (`framework`, `image`, `num_nodes`, `device`, `device_count`, `trainer_type`) from `rt.trainer` and `rt.pretrained_model`, aligning with the Kubeflow SDK's `types.Runtime` dataclass.
- Retains backward-compatible fallback for objects with `.spec`.
- Updates `_get_runtime_image` to check `client.get_runtime(name)` first (which resolves both namespaced and cluster-scoped runtimes) before falling back to `CustomObjectsApi`.
- Refactors extraction into `_extract_runtime_metadata` to keep cyclomatic complexity low.
- Adds comprehensive unit tests in `kubeflow_mcp/trainer/api/discovery_test.py`.

Fixes #191

### How it was tested

- `uv run pytest kubeflow_mcp/trainer/api/discovery_test.py` (14 passed)
- `uv run pytest` (516 passed, all unit & conformance schema snapshot tests pass)
- `uv run ruff check .` and `uv run ruff format --check .` (clean)

### DIFF
--- a/kubeflow_mcp/trainer/api/discovery.py
+++ b/kubeflow_mcp/trainer/api/discovery.py
@@ -192,7 +192,7 @@ def get_training_job(name: str, namespace: str | None = None) -> dict[str, Any]:
 
 
 def list_runtimes() -> dict[str, Any]:
-    """List available ClusterTrainingRuntimes.
+    """List available TrainingRuntimes and ClusterTrainingRuntimes.
 
     Call this if ``fine_tune()`` fails with "runtime not found" to see
     what runtimes are installed in the cluster.
@@ -231,17 +231,23 @@ def list_runtimes() -> dict[str, Any]:
         ).model_dump()
 
 
-def _get_runtime_image(name: str) -> str | None:
-    """Extract container image from a ClusterTrainingRuntime CRD or SDK Runtime."""
-    try:
-        client = get_trainer_client()
-        rt = client.get_runtime(name=name)
-        trainer = getattr(rt, "trainer", None)
+def _get_runtime_image(name: str, runtime_obj: Any = None) -> str | None:
+    """Extract container image from an SDK Runtime or ClusterTrainingRuntime CRD."""
+    if runtime_obj is not None:
+        trainer = getattr(runtime_obj, "trainer", None)
         image = getattr(trainer, "image", None) if trainer is not None else None
         if image:
             return str(image)
-    except Exception:
-        pass
+    else:
+        try:
+            client = get_trainer_client()
+            rt = client.get_runtime(name=name)
+            trainer = getattr(rt, "trainer", None)
+            image = getattr(trainer, "image", None) if trainer is not None else None
+            if image:
+                return str(image)
+        except Exception:
+            pass
 
     try:
         api = get_custom_objects_api()
@@ -268,7 +274,7 @@ def _get_runtime_image(name: str) -> str | None:
     return None
 
 
-def _fetch_packages_via_pod(runtime_name: str) -> dict[str, Any]:
+def _fetch_packages_via_pod(runtime_name: str, runtime_obj: Any = None) -> dict[str, Any]:
     """Create a lightweight Pod to run ``pip list`` using the runtime's image.
 
     Creates a temporary Pod from the runtime's container image to list packages.
@@ -277,7 +283,7 @@ def _fetch_packages_via_pod(runtime_name: str) -> dict[str, Any]:
     The pod includes emptyDir volumes for writable directories so it works
     on platforms with read-only root filesystems.
     """
-    image = _get_runtime_image(runtime_name)
+    image = _get_runtime_image(runtime_name, runtime_obj=runtime_obj)
     if not image:
         return {
             "packages_error": f"Could not extract container image from runtime '{runtime_name}'"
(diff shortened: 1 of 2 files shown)


kubeflow_mcp/trainer/api/discovery.py
```diff
@@ -360,6 +373,62 @@ def _fetch_packages_via_pod(runtime_name: str) -> dict[str, Any]:
...
+    trainer = getattr(rt, "trainer", None)
+    if trainer is not None:
+        framework = getattr(trainer, "framework", None)
+        if framework:
+            data["framework"] = framework
+        image = getattr(trainer, "image", None)
+        if image:
+            data["image"] = str(image)
+        num_nodes = getattr(trainer, "num_nodes", None)
+        if num_nodes:
```
abhijeet-dhumal:
```suggestion
        if num_nodes is not None:
```
`if num_nodes` skips when num_nodes is 0, because 0 is falsy in Python.
But 0 is a valid value so the response drops num_nodes entirely.

kubeflow_mcp/trainer/api/discovery_test.py
```diff
@@ -104,12 +105,125 @@ def test_rejects_invalid_resource_name_before_calling_sdk(tool, kwargs, client_p
...
-# TODO(test): list_training_jobs - namespace policy enforcement
-# TODO(test): get_training_job - returns job details with mock SDK
-# TODO(test): list_runtimes - returns runtime list
-# TODO(test): get_runtime - returns runtime details
-# TODO(test): get_runtime - include_packages=True spawns pod
-# TODO(test): get_runtime - not found returns RESOURCE_NOT_FOUND
+class TestGetRuntime:
+    @patch("kubeflow_mcp.trainer.api.discovery.get_trainer_client")
+    def test_get_runtime_extracts_sdk_trainer_metadata(self, mock_client_fn):
+        mock_trainer = MagicMock()
```
abhijeet-dhumal, line 111: Thinking would it be good to add a test with real Runtime / RuntimeTrainer objects instead of MagicMock so this doesn't slip through again : https://github.com/kubeflow/mcp-server/pull/192/changes#r3901070952

kubeflow_mcp/trainer/api/discovery.py
```diff
@@ -360,6 +373,62 @@ def _fetch_packages_via_pod(runtime_name: str) -> dict[str, Any]:
...
+                            None,
+                        ),
+                    }
+                    for rj in replicated_jobs
+                ]
+
+    return data
+
+
 def get_runtime(name: str, include_packages: bool = False) -> dict[str, Any]:
```
abhijeet-dhumal, line 438: include_packages=True triggers a second client.get_runtime() via _get_runtime_image(). Thinking .. we already have rt in get_runtime().. can we pass trainer.image through (or the whole rt) to avoid duplicate API calls on the slow path. WDYT?

kubeflow_mcp/trainer/api/discovery.py
```diff
@@ -360,6 +373,62 @@ def _fetch_packages_via_pod(runtime_name: str) -> dict[str, Any]:
...
     """Get ClusterTrainingRuntime configuration.
```
abhijeet-dhumal: Docstring still says "ClusterTrainingRuntime" only. SDK resolves both cluster-scoped and namespaced runtimes right?.. worth updating while this is not merged yet..

priyank766:
🚀 Thanks for the great feedback and suggestions @abhijeet-dhumal! Updated:                                  
                                                                                                              
1. Changed `if num_nodes:` to `if num_nodes is not None:` so `0` is preserved.                            
2. Passed the resolved `rt` object into `_fetch_packages_via_pod()` and `_get_runtime_image()` to avoid duplicate API calls.                                                                                        
3. Updated `get_runtime()` and `list_runtimes()` docstrings to mention both `TrainingRuntime` and `ClusterTrainingRuntime`.                                                                                   
4. Added unit tests using real SDK `Runtime` and `RuntimeTrainer`.

abhijeet-dhumal:
Thanks @priyank766 !
/lgtm
/approve

PR #196, opened 2026-09-02
feat(test): add negative path and lifecycle scenarios to Kubernetes E2E suite

### LINKED ISSUE #169: feat(tests): add negative path and lifecycle scenarios to Kubernetes E2E suite
## Summary

Extend the E2E test suite (#109) with negative paths, confirm gate enforcement, suspend/resume lifecycle, monitoring, and platform tool coverage against a real Kind cluster.

## Why

#109 proves tools work with valid inputs. It does not exercise error boundaries, confirm gate blocking, or the full job lifecycle — the top failure modes agents hit in practice.

## Scope

- Negative paths — invalid K8s name, non-existent namespace/job, fine_tune on GPU-less cluster
- Confirm gate — confirmed=False blocks mutation, returns preview, no CR created
- Lifecycle — suspend, resume, delete (note: suspended jobs show status "Created" in the API, known controller behavior)
- Monitoring — get_training_events on a live job, get_training_logs (assert success only, pod may not reach Running)
- Discovery — get_runtime(name) for an installed runtime
- Platform — inspect_crd() list and detail, inspect_controller(), health_check
- Multi-namespace — create a temporary namespace, run a job there, verify namespace param works, clean up

Out of scope (not reliable on Kind):
- estimate_resources — requires network call to HuggingFace Hub, flake-prone in CI
- wait_for_training — blocks MCP connection, timing unpredictable on Kind
- run_container_training — image pull dependency

## Design Constraints

- Independent test functions (no cascade failures)
- Same KUBEFLOW_MCP_E2E=true gate, no new infra beyond #109's make test-e2e-setup-cluster
- Controller is running (inspect_controller feasible)
- CRDs and runtimes pre-installed (inspect_crd, get_runtime feasible)
- Persona/auth out of scope (transport-level concern)

## References

- [PR #109](https://github.com/kubeflow/mcp-server/pull/109) — initial E2E suite
- [Testing MCP Servers: Five Gates](https://dev.to/aws-heroes/testing-mcp-servers-the-five-gates-between-demo-and-production-2inf) — scenario layer guidance
- Related: #63

/cc @prakharsingh-74

priyank766:
## Description

Extends the Kubernetes E2E test suite (`tests/e2e/test_kubernetes_e2e.py`) to cover error boundaries, confirm gate enforcement, job lifecycle, monitoring, discovery, platform tools, and multi-namespace scenarios against a live Kind cluster.

Key changes:
- **Test Independence:** Split into 8 standalone `@pytest.mark.asyncio` test cases so failures don't cascade across unrelated scenarios.
- **Negative Paths:** Added assertions for invalid Kubernetes resource names, non-existent jobs/runtimes, invalid lifecycle actions, and GPU-less fine-tuning validation.
- **Confirm Gate:** Validated that `confirmed=False` returns previews for creation (`run_custom_training`) and deletion (`delete_training_job`) without altering cluster state.
- **Job Lifecycle:** Verified the full suspend -> resume -> delete cycle on live TrainJobs.
- **Monitoring Tools:** Added tests for `get_training_events` and `get_training_logs` on active jobs.
- **Discovery & Platform:** Added tests for `get_runtime` metadata extraction, `health_check`, `inspect_crd` (catalog and detail), and `inspect_controller`.
- **Multi-Namespace Isolation:** Verified TrainJob submission, retrieval, and deletion inside a temporary namespace, ensuring jobs are isolated from the `default` namespace.
- **Resource Hygiene:** Added `try ... finally` blocks across all mutating tests to guarantee created jobs and namespaces are cleaned up on both test success and failure.

## Related Issue

Fixes #169

## Checklist

## Testing

- [x] Unit tests
- [x] Integration tests
- [x] E2E tests

**Testing details:**
- Ran `uv run pytest` (520 passed, all 8 E2E test cases cleanly collected and skipped when `KUBEFLOW_MCP_E2E=true` is unset).
- Ran `uv run ruff check .` and `uv run ruff format --check .` (clean, zero lint or formatting issues).
- Verified that all created jobs and namespaces use unique UUID suffixes and have `try ... finally` cleanup handlers.

### DIFF
--- a/tests/e2e/test_kubernetes_e2e.py
+++ b/tests/e2e/test_kubernetes_e2e.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import json
+import logging
 import os
 import uuid
 from collections.abc import AsyncGenerator
@@ -27,6 +28,8 @@ from mcp import ClientSession, StdioServerParameters
 from mcp.client.stdio import stdio_client
 
 from kubeflow_mcp.common.utils import get_core_v1_api
+
+logger = logging.getLogger(__name__)
 
 # Mark all tests in this file as requiring KUBEFLOW_MCP_E2E=true
 pytestmark = pytest.mark.skipif(
@@ -41,6 +44,11 @@ NOOP_TRAIN_SCRIPT = (
     "if __name__ == '__main__':\n"
     "    train()\n"
 )
+
+INVALID_K8S_NAME = "INVALID_UPPERCASE_NAME"
+NON_EXISTENT_JOB = "non-existent-e2e-job-999"
+NON_EXISTENT_RUNTIME = "non-existent-runtime-999"
+NON_EXISTENT_NAMESPACE = "non-existent-ns-999"
 
 
 @pytest.fixture
@@ -107,6 +115,18 @@ async def _get_custom_runtime_name(session: ClientSession) -> str:
     pytest.skip("No custom torch- training runtime found in the cluster")
 
 
+async def _get_torchtune_runtime_name(session: ClientSession) -> str:
+    """Discover installed torchtune runtime in the cluster."""
+    data = await _call_tool(session, "list_runtimes")
+    assert data.get("success") is True, f"list_runtimes failed: {data}"
+    runtimes = data.get("data", {}).get("runtimes", [])
+    for r in runtimes:
+        name = r.get("name", "")
+        if "torchtune" in name or "torch-tune" in name:
+            return name
+    pytest.skip("No torchtune runtime found in the cluster")
+
+
 @pytest.mark.asyncio
 async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
     """Validate full flow of read-only and mutating trainer MCP tools against a live cluster."""
@@ -140,14 +160,7 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
     assert len(runtimes) > 0
 
     # Pick runtime for fine_tune preview (requires torchtune)
-    torchtune_runtime_name = None
-    for r in runtimes:
-        name = r.get("name", "")
-        if "torchtune" in name or "torch-tune" in name:
-            torchtune_runtime_name = name
-            break
-    if not torchtune_runtime_name:
-        pytest.skip("No torchtune runtime found in the cluster")
+    torchtune_runtime_name = await _get_torchtune_runtime_name(mcp_session)
 
     # Pick runtime for run_custom_training
     custom_runtime_name = await _get_custom_runtime_name(mcp_session)
@@ -242,19 +255,17 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
                 arguments={"name": custom_job_name, "namespace": namespace, "confirmed": True},
             )
         except Exception:
-            pass
+            logger.warning("cleanup failed for %s", custom_job_name, exc_info=True)
 
 
 @pytest.mark.asyncio
 async def test_e2e_negative_paths(mcp_session: ClientSession) -> None:
     """Validate error boundaries and input validation against live server."""
-    invalid_name = "INVALID_UPPERCASE_NAME"
-
     # 1. Invalid K8s names rejected before K8s API call
     resp = await _call_tool(
         mcp_session,
         "get_training_job",
-        {"name": invalid_name, "namespace": "default"},
+        {"name": INVALID_K8S_NAME, "namespace": "default"},
     )
     assert resp.get("success") is False
     assert resp.get("error_code") == "VALIDATION_ERROR"
@@ -262,7 +273,7 @@ async def test_e2e_negative_paths(mcp_session: ClientSession) -> None:
     resp = await _call_tool(
         mcp_session,
         "run_custom_training",
-        {"script": NOOP_TRAIN_SCRIPT, "name": invalid_name},
+        {"script": NOOP_TRAIN_SCRIPT, "name": INVALID_K8S_NAME},
     )
     assert resp.get("success") is False
     assert resp.get("error_code") == "VALIDATION_ERROR"
@@ -270,7 +281,7 @@ async def test_e2e_negative_paths(mcp_session: ClientSession) -> None:
     resp = await _call_tool(
         mcp_session,
         "delete_training_job",
-        {"name": invalid_name, "confirmed": True},
+        {"name": INVALID_K8S_NAME, "confirmed": True},
     )
     assert resp.get("success") is False
     assert resp.get("error_code") == "VALIDATION_ERROR"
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
Thanks @priyank766 !
First of all,.. to validate this PR we need to enable E2E test pr check,
E2E is SKIPPED because ok-to-test-e2e label addition isn't practical.. Prow bot won't allow us apply it via /label.
Can you drop the job if: label check in e2e-k8s.yaml? Path filter is enough. Then this PR should run E2E on sync.
Just need to remove this if block :
```
 if: >-
      github.event_name != 'pull_request' ||
      contains(github.event.pull_request.labels.*.name, 'ok-to-test-e2e')
```

tests/e2e/test_kubernetes_e2e.py
```diff
@@ -174,65 +186,508 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
...
-    assert data["data"]["name"] == custom_job_name
-    assert data["data"]["status"] == "Created"
+    assert resp.get("success") is False
+    assert resp.get("error_code") == "VALIDATION_ERROR"
 
-    # 9. delete_training_job(confirmed=True) removes it, verify gone
-    resp = await mcp_session.call_tool(
-        "delete_training_job",
-        arguments={"name": custom_job_name, "namespace": namespace, "confirmed": True},
+    # 2. Non-existent job returns RESOURCE_NOT_FOUND
```
abhijeet-dhumal, line 289: can we add case for non-existant namespace too ?

tests/e2e/test_kubernetes_e2e.py
```diff
@@ -174,65 +186,508 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
...
+        "get_training_job",
+        {"name": "non-existent-e2e-job-999", "namespace": "default"},
     )
-    assert not resp.isError
-    assert len(resp.content) == 1
-    data = json.loads(resp.content[0].text)
-    assert data["success"] is True
-    assert data["data"]["deleted"] is True
+    assert resp.get("success") is False
+    assert resp.get("error_code") == "RESOURCE_NOT_FOUND"
```
abhijeet-dhumal, line 296: Instead of hard string please use constant here

tests/e2e/test_kubernetes_e2e.py
```diff
@@ -174,65 +186,508 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
...
+    # 5. fine_tune on GPU-less cluster returns VALIDATION_ERROR
+    cluster_res = await _call_tool(mcp_session, "get_cluster_resources")
+    if cluster_res.get("data", {}).get("gpu_total", 0) == 0:
+        resp = await _call_tool(
+            mcp_session,
+            "fine_tune",
+            {
+                "model": "hf://google/gemma-2b",
+                "dataset": "hf://tatsu-lab/alpaca",
+                "runtime": "torch-distributed",
```
abhijeet-dhumal: Nit: runtime is hardcoded to "torch-distributed". Works on CPU-only Kind because GPU validation runs before runtime checks (_check_gpu_available at confirmed=False), but for consistency can we consider reusing a discovered torchtune runtime from list_runtimes same as test_kubernetes_e2e_flow ?

tests/e2e/test_kubernetes_e2e.py
```diff
@@ -174,65 +186,508 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
...
+            "get_training_events",
+            {"name": job_name, "namespace": namespace},
+        )
+        assert events_resp.get("success") is True
+        data = events_resp.get("data", {})
+        assert data.get("job") == job_name
+        assert isinstance(data.get("events"), list)
+        assert data.get("total") >= 0
+
+        # 2. get_training_logs (assert success only, pod may not reach Running in time)
```
abhijeet-dhumal: Potentially on slow Kind: if the pod hasn't scheduled yet, get_training_logs may return success=False rather than empty logs...
wdyt?

tests/e2e/test_kubernetes_e2e.py
```diff
@@ -174,65 +186,508 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
...
+        del_resp = await _call_tool(
+            mcp_session,
+            "delete_training_job",
+            {"name": job_name, "namespace": temp_ns, "confirmed": True},
+        )
+        assert del_resp.get("success") is True
+        assert del_resp.get("data", {}).get("deleted") is True
+    finally:
+        try:
+            core.delete_namespace(name=temp_ns, grace_period_seconds=0)
```
abhijeet-dhumal, line 731: finally deletes the namespace but not the TrainJob if the test fails after job creation. Namespace delete usually cascades, but still.. can we use explicit job cleanup first to avoid any leaks or flakes on re-runs?

tests/e2e/test_kubernetes_e2e.py
```diff
@@ -174,65 +186,508 @@ async def test_kubernetes_e2e_flow(mcp_session: ClientSession) -> None:
...
+        assert data["success"] is False
+        assert data["error_code"] == "RESOURCE_NOT_FOUND"
+    finally:
+        try:
+            await mcp_session.call_tool(
+                "delete_training_job",
+                arguments={"name": custom_job_name, "namespace": namespace, "confirmed": True},
+            )
+        except Exception:
+            pass
```
abhijeet-dhumal: Nit: except Exception: pass in teardown hides cleanup failures. 
Consider logging.warning("cleanup failed for %s", job_name, exc_info=True).. 
this will help us debug CI leaks ..

tests/e2e/test_kubernetes_e2e.py
priyank766: Good catch! Added a negative test case for non-existent namespace as well.

tests/e2e/test_kubernetes_e2e.py
priyank766: Defined module constants for these and updated all the calls!

tests/e2e/test_kubernetes_e2e.py
priyank766: Makes sense! Added a helper to discover the torchtune runtime from list_runtimes and reused it here as well.

tests/e2e/test_kubernetes_e2e.py
priyank766: Really good catch! Updated the assertion so that if the pod is still pending scheduling on slow runners, it accepts SDK_ERROR / RESOURCE_NOT_FOUND rather than flaking.

tests/e2e/test_kubernetes_e2e.py
priyank766: Agreed! Added explicit TrainJob deletion before namespace deletion in the finally block.

tests/e2e/test_kubernetes_e2e.py
priyank766: Great point! Replaced all `except Exception: pass` teardowns with `logger.warning(..., exc_info=True)` so any cleanup issues are logged.

priyank766:
🚀Thanks a lot for the review and suggestions @abhijeet-dhumal!

I've addressed all the feedback and pushed the changes. Dropped the workflow label check as well, so the Kind E2E check is now running directly on sync.

Please take a look when you have a moment!

abhijeet-dhumal:
Thanks @priyank766, great work ! 🚀 
/lgtm
/approve

PR #198, opened 2026-09-02
feat(trainer): add confirm gate to update_training_job

### LINKED ISSUE #121: add confirm gate to update_training_job
### Description

`update_training_job` is the only mutating trainer tool that still runs immediately. It has no `confirmed` parameter.

Other write tools preview first (`confirmed=False`) and only apply with `confirmed=True`. AGENTS.md already calls this a legacy exception. We should bring `update_training_job` in line with that pattern.

### Use Case

Suspend/resume should use the same preview-before-mutate flow as delete and job submit, so agents (and humans reviewing them) are not surprised.

### Component

TrainerClient

### References

- `delete_training_job` (pattern to follow)

priyank766:
## Description

Brings `update_training_job` in line with the standard confirm-gate pattern used across all mutating tools in the server.

### Key Changes
- **Tool Confirm Gate**: Added `confirmed: bool = False` to `update_training_job`. Calling with `confirmed=False` returns a `PreviewResponse` containing the action, target job, and namespace. Mutation executes only when `confirmed=True`.
- **Security & Ownership Order**: Placed the preview gate after namespace validation and MCP ownership checks (`is_mcp_managed`), ensuring unauthorized personas cannot obtain preview responses for unowned jobs.
- **Documentation & Instructions**: 
  - Updated `CLIENT_TOOL_DESCRIPTIONS` and `INSTRUCTION_SECTIONS` in `trainer/__init__.py`.
  - Updated recovery examples in `troubleshooting.md`.
  - Removed "legacy exception" wording from `AGENTS.md`, `README.md`, and `docs/CONVENTIONS.md`.
- **Conformance Snapshot**: Updated `tests/conformance/snapshots/tool_schema_platform_admin.json` to reflect the new `confirmed` parameter.
- **Tests**:
  - Added unit test suite in `kubeflow_mcp/trainer/api/lifecycle_test.py` covering validation, preview responses, non-admin ownership enforcement, confirmed suspend/resume, and 404 handling.
  - Updated `test_e2e_confirm_gate` and `test_e2e_job_lifecycle` in `tests/e2e/test_kubernetes_e2e.py` to verify preview and confirmed execution against a live cluster.

## Related Issue

Fixes #121

## Testing

- Unit tests: `uv run pytest kubeflow_mcp/trainer/api/lifecycle_test.py` (13 passed)
- Full test suite: `uv run pytest` (529 passed, 0 failures)
- Schema snapshot: `uv run pytest tests/conformance/tool_schema_snapshot_test.py` (passed)
- Linting & formatting: `uv run ruff check .` and `uv run ruff format --check .` (clean)

### DIFF
--- a/kubeflow_mcp/trainer/api/lifecycle_test.py
+++ b/kubeflow_mcp/trainer/api/lifecycle_test.py
@@ -20,9 +20,20 @@ marked as TODOs.
 
 from __future__ import annotations
 
-import pytest
-from tests.common import FAILED, VALIDATION_ERROR, TestCase, assert_test_case
+from unittest.mock import MagicMock, patch
 
+import pytest
+from kubeflow.trainer.constants import constants as trainer_constants
+from kubernetes.client.exceptions import ApiException
+from tests.common import (
+    FAILED,
+    RESOURCE_NOT_FOUND,
+    VALIDATION_ERROR,
+    TestCase,
+    assert_test_case,
+)
+
+from kubeflow_mcp.common import utils as mcp_utils
 from kubeflow_mcp.trainer.api.lifecycle import delete_training_job, update_training_job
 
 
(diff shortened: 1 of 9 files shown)


kubeflow_mcp/trainer/resources/troubleshooting.md
abhijeet-dhumal: Missed this reference.. should include confirmed=True to match the updated signature

kubeflow_mcp/trainer/api/lifecycle_test.py
```diff
@@ -47,11 +58,35 @@ def test_delete_training_job_validation(test_case):
     assert_test_case(test_case, delete_training_job)
 
 
-# TODO(test): test preview (confirmed=False) returns job details
-# TODO(test): test confirmed=True with mock SDK deletes job
-# TODO(test): test not found returns RESOURCE_NOT_FOUND
-# TODO(test): test namespace policy enforcement
-# TODO(test): test non-admin persona cannot delete non-MCP jobs
+class TestDeleteTrainingJob:
```
abhijeet-dhumal, line 64: The update-side coverage is solid. Can you also add the delete-side tests while you're here? Three cases are still uncovered: delete_training_job with a 404 returns RESOURCE_NOT_FOUND, namespace policy enforcement blocks restricted namespaces, and non-admin persona cannot delete a non-MCP job. The update tests you wrote (test_confirmed_not_found_returns_resource_not_found, test_non_admin_cannot_preview_non_mcp_job) are the exact pattern to follow.. should be a quick copy-adapt.

kubeflow_mcp/trainer/resources/troubleshooting.md
priyank766: Good catch! Updated the reference to include `confirmed=True`.

kubeflow_mcp/trainer/api/lifecycle_test.py
priyank766: Added tests for all three delete cases (404, namespace policy, and non-admin ownership). Thanks!

abhijeet-dhumal:
Thanks @priyank766, Great work! 🚀 
/lgtm
/approve

PR #204, opened 2026-09-05
chore: added unit tests for health.py tools and metadata

### LINKED ISSUE #203: test(chore): add unit tests for health.py tools and metadata
Part of https://github.com/kubeflow/mcp-server/issues/9. To build on the test scaffold in https://github.com/kubeflow/mcp-server/pull/6.

Raise core module coverage and clear TODO(test) markers under tests/unit/core/

Currently, `kubeflow_mcp/core/health.py ` don't have any separate unit test for it. It is covered under e2e tests, but unit test are needed to check whether each function on this file is working or not.

PROPOSED CHANGES:
- `get_server_logs` : input validation (`limit < 1`), log filtering by level and error handling
Add `kubeflow_mcp/core/health_test.py` covering:
- `health_check` : healthy state (mocked k8s API success) and degraded state 
- Metadata integrity for `HEALTH_TOOLS` , descriptions, and annotations.

modichika:
## Description

- Adds unit test for `kubeflow_mcp/core/health.py`
- Covers `health_check()` healthy(success) and degraded(fail) states.
- Covers `get_server_logs()` validation, log level filtering, and exception handling
- Verifies metadata exports (`HEALTH_TOOLS`, annotations, descriptions)
- All tests pass locally with `make verify`, `make test-python` and `uv run pytest`

## Related Issue

Fixes https://github.com/kubeflow/mcp-server/issues/203

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

### DIFF
--- a/kubeflow_mcp/core/health_test.py
+++ b/kubeflow_mcp/core/health_test.py
@@ -1,0 +1,148 @@
+# Copyright The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for core/health.py - health check and server logs tools."""
+
+from __future__ import annotations
+
+from typing import Any
+from unittest.mock import MagicMock, patch
+
+import pytest
+from tests.common import SDK_ERROR, VALIDATION_ERROR
+
+from kubeflow_mcp.common.utils import K8S_TIMEOUT
+from kubeflow_mcp.core.health import (
+    HEALTH_TOOL_ANNOTATIONS,
+    HEALTH_TOOL_DESCRIPTIONS,
+    HEALTH_TOOLS,
+    get_server_logs,
+    health_check,
+)
+
+
+class TestHealthCheck:
+    """Tests for the health_check tool."""
+
+    @patch("kubeflow_mcp.common.utils.get_core_v1_api")
+    def test_health_check_healthy_when_k8s_succeeds(self, mock_get_core_v1_api: MagicMock) -> None:
+        mock_v1 = MagicMock()
+        mock_get_core_v1_api.return_value = mock_v1
+
+        result = health_check()
+
+        assert result["success"] is True
+        data = result["data"]
+        assert data["status"] == "healthy"
+        assert data["kubernetes"] is True
+        assert isinstance(data["uptime_seconds"], int)
+        assert data["uptime_seconds"] >= 0
+        assert "timestamp" in data
+        mock_v1.list_namespace.assert_called_once_with(limit=1, _request_timeout=K8S_TIMEOUT)
+
+    @patch("kubeflow_mcp.common.utils.get_core_v1_api")
+    def test_health_check_degraded_when_k8s_fails(self, mock_get_core_v1_api: MagicMock) -> None:
+        mock_v1 = MagicMock()
+        mock_v1.list_namespace.side_effect = ConnectionError("K8s API unreachable")
+
+        mock_get_core_v1_api.return_value = mock_v1
+        result = health_check()
+
+        # The tool should not crash even if k8s fails, it should return degraded status
+        assert result["success"] is True
+        data = result["data"]
+        assert data["status"] == "degraded"
+        assert data["kubernetes"] is False
+
+
+class TestGetServerLogs:
+    """Tests for the get_server_logs tool."""
+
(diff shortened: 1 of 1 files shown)


kubeflow_mcp/core/health_test.py
```diff
@@ -0,0 +1,148 @@
...
+    HEALTH_TOOLS,
+    get_server_logs,
+    health_check,
+)
+
+
+class TestHealthCheck:
+    """Tests for the health_check tool."""
+
+    @patch("kubeflow_mcp.common.utils.get_core_v1_api")
```
abhijeet-dhumal, line 38: Did you confirm this patch path works with the lazy import inside health_check()? I guess It should .. but  just checking you didn't hit an import-order surprise locally.

kubeflow_mcp/core/health_test.py
```diff
@@ -0,0 +1,148 @@
...
+                "level": "CRITICAL",
+                "message": "critical message",
+            },
+            {"timestamp": "2026-09-05T00:00:001Z", "level": "INFO", "message": "info message 2"},
+        ]
+
+    def test_reject_limit_less_than_one(self) -> None:
+        for invalid_limit in [0, -1, -50]:
+            result = get_server_logs(limit=invalid_limit)
+            assert result["success"] is False
```
abhijeet-dhumal, line 94: Should we also assert get_log_buffer is never called when limit < 1, to pin the short-circuit before buffer read? WDYT?

kubeflow_mcp/core/health_test.py
```diff
@@ -0,0 +1,148 @@
...
+            assert result["error_code"] == VALIDATION_ERROR
+            assert f"limit must be >= 1, got {invalid_limit}" in result["error"]
+
+    @patch("kubeflow_mcp.core.health.get_log_buffer")
+    def test_filter_by_level_warning(
+        self, mock_get_buffer: MagicMock, sample_logs: list[dict[str, Any]]
+    ) -> None:
+        mock_get_buffer.return_value = sample_logs
+
+        result = get_server_logs(level="WARNING")
```
abhijeet-dhumal, line 104: Would a default level="INFO" case (DEBUG excluded) be worth adding as the most common call path?

kubeflow_mcp/core/health_test.py
```diff
@@ -0,0 +1,148 @@
...
+        result = get_server_logs()
+
+        assert result["success"] is False
+        assert result["error_code"] == SDK_ERROR
+
+
+class TestHealthMetadata:
+    """Tests for metadata, registrations, and annotations."""
+
+    def test_health_tools_exported(self) -> None:
```
abhijeet-dhumal, line 128: This mirrors architecture_test.py, should we extract a shared helper eventually, or is duplication fine at this scale?

abhijeet-dhumal:
@modichika Are you still working on this pr?

kubeflow_mcp/core/health_test.py
```diff
@@ -0,0 +1,148 @@
...
+        assert health_check in HEALTH_TOOLS
+        assert get_server_logs in HEALTH_TOOLS
+        assert len(HEALTH_TOOLS) == 2
+
+    def test_descriptions_defined_for_all_tools(self) -> None:
```
abhijeet-dhumal, line 187: Please add an integration-level assertion using create_server() and list_tools() to verify both health tools are exposed with the expected metadata. Constants alone would not catch a registration regression.

modichika:
Yeah @abhijeet-dhumal sorry caught a fever but I'll make the changes by tmrw. 

Will it be ok?

abhijeet-dhumal:
> Yeah @abhijeet-dhumal sorry caught a fever but I'll make the changes by tmrw.
> 
> Will it be ok?

Hey NP, Pleas take rest and get well soon !

kubeflow_mcp/core/health_test.py
modichika: done!

kubeflow_mcp/core/health_test.py
modichika: i think for just two tools it's too much to create a shared helper at this scale.
we can keep this in duplication files as to address each file at a time.
but definitely a shared helper can be added in the future when more tools are built

kubeflow_mcp/core/health_test.py
modichika: ack!
yes it's worth adding coz DEBUG is for the raw level code, it's for humans to look at as it's complex and may increase tokens of llms if sent to them.
And as our get_server_logs defaults to "INFO" i think adding a test is worth it.

kubeflow_mcp/core/health_test.py
modichika: yes coz the get_log_buffer is very expensive so to call it when limit < 1 makes no sense to me.

kubeflow_mcp/core/health_test.py
modichika: yes it works fine as it's a lazy call.

modichika:
addressed the comments and resolved merging conflicts!

abhijeet-dhumal:
Thank you @modichika!
/lgtm
/approve

PR #210, opened 2026-09-06
fix: report zero lines for empty logs

### LINKED ISSUE #209: fix: report zero lines for empty training logs
Description:

## Problem

The `get_training_logs()` tool reports the number of log lines using:

    "lines": len(sanitized.split("\n"))

When a training job has no logs, `sanitized` is an empty string. In Python:

    "".split("\n") == [""]

As a result, the tool returns `lines: 1` even though no log lines are present.

This makes the response inaccurate for jobs that have not produced logs yet or whose logs are unavailable.

## Proposed Fix

Count empty logs as zero lines:

    "lines": len(sanitized.splitlines()) if sanitized else 0

The existing `follow=True` response should continue returning `lines: 1`, because it contains an informational message.

## Acceptance Criteria

- Empty training logs return `lines: 0`.
- Non-empty logs continue to report the correct line count.
- The `follow=True` response remains unchanged.
- Add a focused unit test covering the empty-log case.

pratik-naik003:
## Description

Fixes the line count returned by `get_training_logs()` when a training job has no logs.

Previously, the tool used:

    "lines": len(sanitized.split("\n"))

For an empty string, Python returns one empty element, causing the response to incorrectly report `lines: 1`.

The implementation now returns `lines: 0` for empty logs while preserving the existing behavior for non-empty logs and `follow=True`.

## Related Issue

Fixes #209

## Changes

- Updated empty-log line counting in `kubeflow_mcp/trainer/api/monitoring.py`.
- Added a focused regression test.
- Preserved existing behavior for non-empty and streaming responses.

## Testing

- [x] `uv run pytest -q kubeflow_mcp/trainer/api/monitoring_test.py` — 48 passed
- [x] Ruff checks passed
- [x] Ruff formatting checks passed
- [x] Commit signed off


abhijeet-dhumal:
Thanks @pratik-naik003, good catch!
/lgtm

PR #215, opened 2026-09-07
fix: validate server log levels

### LINKED ISSUE #214: fix: validate log level in get_server_logs
## Problem

The `get_server_logs()` tool accepts a `level` argument, but invalid log levels are silently treated as `INFO`.

For example:

    get_server_logs(level="WARN")

The supported levels are `DEBUG`, `INFO`, `WARNING`, `ERROR`, and `CRITICAL`. However, the implementation uses:

    min_level = level_order.get(level.upper(), 1)

If the provided value is invalid or contains a typo, the tool defaults to INFO logs instead of returning a clear validation error. This can cause users to receive unexpected log results without realizing that their input was incorrect.

## Proposed Fix

Validate the requested log level before filtering logs.

- Accept only `DEBUG`, `INFO`, `WARNING`, `ERROR`, and `CRITICAL`.
- Return a `VALIDATION_ERROR` for unsupported values.
- Keep the existing default `INFO` behavior when the argument is omitted.
- Do not read or filter the log buffer when validation fails.

## Acceptance Criteria

- Valid log levels continue to work.
- Invalid values such as `WARN` or `INVALID` return `VALIDATION_ERROR`.
- The default `level="INFO"` behavior remains unchanged.
- Focused unit tests cover valid, invalid, and default level values.

pratik-naik003:
## Description

Adds validation for the `level` argument in `get_server_logs()`.

Previously, unsupported values such as `WARN` or `INVALID` were silently treated as `INFO`, which could return unexpected log results without notifying the caller.

## Changes

- Validate log levels before reading the log buffer.
- Return `VALIDATION_ERROR` for unsupported values.
- Preserve case-insensitive valid levels such as `error`.
- Document the supported `CRITICAL` level.
- Added focused unit tests for invalid and valid log levels.

## Related Issue

Fixes #214

## Testing

- [x] `uv run pytest -q kubeflow_mcp/core/health_test.py` — 2 passed
- [x] Ruff checks passed
- [x] Ruff formatting checks passed
- [x] Commit signed off


kubeflow_mcp/core/health.py
```diff
@@ -91,7 +91,13 @@ def get_server_logs(
             "ERROR": 3,
             "CRITICAL": 4,
         }
-        min_level = level_order.get(level.upper(), 1)
+        normalized_level = level.upper()
+        if normalized_level not in level_order:
```
abhijeet-dhumal, line 99: Please normalize WARN/FATAL or reuse the logging level mapping. setup_logging() accepts these values, but get_server_logs() rejects them, making valid LOG_LEVEL configurations fail at runtime.

pratik-naik003:
@abhijeet-dhumal  okay as you said i did the changes check now

abhijeet-dhumal:
Thanks @pratik-naik003, good catch!
/ok-to-test

PR #220, opened 2026-09-07
fix(core): keep the tail when truncating log output

### LINKED ISSUE #219: fix(core): log truncation drops the end of the log, where the failure is
## Problem

`get_training_logs()` builds its response in two steps (`kubeflow_mcp/trainer/api/monitoring.py`):

```python
log_lines = list(deque(client.get_job_logs(...), maxlen=MAX_LOG_LINES))
...
logs = "\n".join(log_lines)
sanitized = truncate_log_output(logs)
```

The `deque` keeps the **last** `MAX_LOG_LINES` (1000) lines, which is what you want when a job has failed. `truncate_log_output()` then keeps the **first** 10000 characters of that window (`kubeflow_mcp/core/security.py`):

```python
output = output[:max_length] + f"\n... (truncated, {len(output)} total chars)"
```

1000 log lines are almost always more than 10000 characters, so what comes back is the oldest part of the tail window -- pip installs, shard downloads, and the traceback at the end is dropped.

The response can then contradict itself. `extract_failure_hint()` runs on the full text, so `failure_hint` reports the OOM while `logs` shows nothing failing.

## Reproduction

```python
from unittest.mock import patch, MagicMock
from kubeflow_mcp.trainer.api.monitoring import get_training_logs

lines = [f"[{i:05d}] downloading shard {i}/900 ..." for i in range(900)]
lines.append("torch.cuda.OutOfMemoryError: CUDA out of memory")

with (
    patch("kubeflow_mcp.trainer.api.monitoring.check_namespace_allowed", return_value=None),
    patch("kubeflow_mcp.trainer.api.monitoring.get_trainer_client_for_namespace") as gc,
):
    client = MagicMock()
    client.get_job_logs.return_value = iter(lines)
    gc.return_value = client
    data = get_training_logs("demo-job")["data"]

print(data["failure_hint"]["category"])
print("OutOfMemoryError" in data["logs"])
```

```
OOM
False
```

`test_log_truncation` in `monitoring_test.py` does not catch this. Its 1000-line fixture joins to 9499 characters, just under the 10000 limit, so the character branch of `truncate_log_output()` never runs in the suite.

## Proposed Fix

Keep the tail, and move the marker to the front so the reader still knows content was dropped:

```python
if len(output) > max_length:
    output = f"... (truncated, {len(output)} total chars)\n" + output[-max_length:]
```

`get_training_logs()` is the only caller and already passes the most recent window, so no call site changes.

## Acceptance Criteria

- A failure signature at the end of a long log is present in the returned `logs`.
- `failure_hint` and `logs` no longer disagree about what the job printed.
- Output shorter than `max_length` is returned unchanged.
- Tests cover the direction of truncation, both on the helper and through `get_training_logs()`.

<br/>

**P.S.**: I'll shortly open a PR addressing this. Thanks!

Neilblaze:
## Description

`get_training_logs()` keeps the last `MAX_LOG_LINES` lines with a `deque`, then hands that window to `truncate_log_output()`, which cut it down to its first 10000 characters. A 1000-line window is nearly always longer than that, so the traceback at the end was dropped and the caller got the shard downloads instead.

The response could also contradict itself: `extract_failure_hint()` reads the full text, so `failure_hint` reported `OOM` on a `logs` value containing no sign of it.

`truncate_log_output()` now keeps the tail and puts the truncation marker at the front. `get_training_logs()` is the only caller and already passes the most recent window, so no call site changed. `truncate_log_output()` also rejects a `max_length` below 1. A zero limit sliced as `output[-0:]`, which is the whole log.
ㅤ

Four tests, all of which fail on `main`:

- `test_truncate_keeps_tail` pins the direction on the helper.
- `test_failure_at_end_survives_char_truncation` drives it through `get_training_logs()` with an OOM on the last line, and asserts `failure_hint` and `logs` agree.
- `test_truncate_rejects_non_positive_max_length` checks that `0` and `-1` raise instead of returning the whole log.
- `test_truncate_smallest_max_length_keeps_last_char` pins the smallest valid limit.

Worth noting why this was not caught: the existing `test_log_truncation` covers the line cap, not the character cap -- its fixture joins to 9499 characters, just under the 10000 limit.
ㅤ

## Related Issue
Fixes #219

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

> [!NOTE]
> `566 passed, 8 skipped`; `ruff check` / `ruff format --check` / `pre-commit run --all-files` clean; tool schema snapshot unchanged.

Ran the reproduction from the issue before and after. Before: `failure_hint` was `OOM` while `"OutOfMemoryError" in logs` was `False`. After: both print as expected and the response opens with the `... (truncated, N total chars)` marker.

### DIFF
--- a/kubeflow_mcp/core/security.py
+++ b/kubeflow_mcp/core/security.py
@@ -206,9 +206,14 @@ def is_safe_python_code(code: str) -> tuple[bool, str]:
 
 
 def truncate_log_output(output: str, max_length: int = 10000) -> str:
-    """Truncate log output to a safe length for MCP responses."""
+    """Truncate log output to a safe length for MCP responses.
+
+    Keeps the tail. Callers pass the most recent slice of a workload's log,
+    where a failure lands, so cutting from the front drops the part worth
+    reading.
+    """
     if len(output) > max_length:
-        output = output[:max_length] + f"\n... (truncated, {len(output)} total chars)"
+        output = f"... (truncated, {len(output)} total chars)\n" + output[-max_length:]
     return output
 
 
--- a/kubeflow_mcp/core/security_test.py
+++ b/kubeflow_mcp/core/security_test.py
@@ -372,4 +372,11 @@ def test_truncate_long_output():
     assert "truncated" in result
 
 
+def test_truncate_keeps_tail():
+    output = "start-of-log\n" + "filler\n" * 500 + "fatal error at the end"
+    result = truncate_log_output(output, max_length=100)
+    assert result.endswith("fatal error at the end")
+    assert "start-of-log" not in result
+
+
 # TODO(test): test exact boundary (max_length characters)
--- a/kubeflow_mcp/trainer/api/monitoring_test.py
+++ b/kubeflow_mcp/trainer/api/monitoring_test.py
@@ -254,6 +254,17 @@ class TestGetTrainingLogs:
 
     @patch(PATCH_NS_CHECK, return_value=None)
     @patch(PATCH_CLIENT)
+    def test_failure_at_end_survives_char_truncation(self, mock_client_fn, _ns):
+        lines = [f"[{i:05d}] downloading shard {i}" for i in range(900)]
+        lines.append("torch.cuda.OutOfMemoryError: CUDA out of memory")
+        mock_client_fn.return_value = _make_mock_client(get_job_logs=lines)
+        result = get_training_logs("oom-at-end-job")
+        assert result["success"] is True
+        assert result["data"]["failure_hint"]["category"] == "OOM"
+        assert "OutOfMemoryError" in result["data"]["logs"]
+
+    @patch(PATCH_NS_CHECK, return_value=None)
+    @patch(PATCH_CLIENT)
     def test_log_truncation_with_generator(self, mock_client_fn, _ns):
         def _log_generator():
             for i in range(MAX_LOG_LINES + 500):


Neilblaze:
CC'ing @abhijeet-dhumal for greater visibility. Thanks!

kubeflow_mcp/core/security.py
```diff
@@ -206,9 +206,14 @@ def is_safe_python_code(code: str) -> tuple[bool, str]:
...
-    """Truncate log output to a safe length for MCP responses."""
+    """Truncate log output to a safe length for MCP responses.
+
+    Keeps the tail. Callers pass the most recent slice of a workload's log,
+    where a failure lands, so cutting from the front drops the part worth
+    reading.
+    """
     if len(output) > max_length:
-        output = output[:max_length] + f"\n... (truncated, {len(output)} total chars)"
+        output = f"... (truncated, {len(output)} total chars)\n" + output[-max_length:]
```
abhijeet-dhumal, line 219: With max_length=0, output[-0:] returns the entire log, defeating the size limit. Please reject non-positive values or special-case zero, and add a boundary test.

Neilblaze:
@abhijeet-dhumal PTAL! :) ... I think you can now trigger `/ok-to-test`

Neilblaze:
@abhijeet-dhumal All green ... I think we're gtg! :)

abhijeet-dhumal:
/hold
Please fix the zero/negative max_length regression and add boundary tests.

PR #222, opened 2026-09-07
fix(core): keep the circuit breaker from wedging in half-open

### LINKED ISSUE #221: fix(core): circuit breaker can get stuck half-open and never recover
## Problem

`CircuitBreaker.can_execute()` hands out at most `half_open_max_calls` probe slots and never takes one back (`kubeflow_mcp/core/resilience.py`):

```python
if self.half_open_calls < self.half_open_max_calls:
    self.half_open_calls += 1
    return True
return False
```

`record_success()` needs that same number of successes to close the circuit. So every probe has to report an outcome, or a slot is gone for good.

`_audit_wrap` does not always report one (`kubeflow_mcp/core/server.py`):

```python
if is_success:
    breaker.record_success()
elif is_infrastructure_error(result):
    breaker.record_failure()
```

`is_infrastructure_error()` only matches `KUBERNETES_ERROR`, `SDK_ERROR`, and `TIMEOUT`, so a routine `RESOURCE_NOT_FOUND` or `VALIDATION_ERROR` falls through both branches. The probe ran, the slot is spent, and nothing was recorded.

Once the slots are used up, `can_execute()` returns `False` for every call. Nothing executes, so nothing can report back, so the state never changes — and the `OPEN -> HALF_OPEN` recovery branch cannot help because the state is already `HALF_OPEN`. The tool is dead until the process restarts, while every response says `"Circuit breaker open for '<tool>' — K8s API may be degraded. Retries automatically after recovery timeout."`

`dynamic_tools.execute_tool()` already does the opposite for the same result:

```python
if isinstance(result, dict) and is_infrastructure_error(result):
    breaker.record_failure()
else:
    breaker.record_success()
```

so `full` mode and `progressive` / `semantic` mode account for the same tool result differently.

Separately, `CircuitBreaker` and `RateLimiter` measure elapsed time with `time.time()`, while `_audit_wrap` already uses `time.monotonic()`. A forward clock step holds the circuit open long past `recovery_timeout`; a backward step drives the limiter's token count negative, and every tool returns `RATE_LIMITED` until it refills.

---

## Reproduction

```python
import time
from unittest.mock import patch, MagicMock
from kubernetes.client.exceptions import ApiException

from kubeflow_mcp.core import resilience
from kubeflow_mcp.core.server import _audit_wrap
from kubeflow_mcp.trainer.api.discovery import get_training_job

resilience.reset_breakers()
resilience.configure_circuit_breaker(failure_threshold=5, recovery_timeout=0.2)
tool = _audit_wrap(get_training_job)

with (
    patch("kubeflow_mcp.trainer.api.discovery.check_namespace_allowed", return_value=None),
    patch("kubeflow_mcp.trainer.api.discovery.get_trainer_client_for_namespace") as gc,
):
    client = MagicMock()
    gc.return_value = client

    # the API is unhealthy for a moment, so the breaker opens as designed
    client.get_job.side_effect = RuntimeError("connection refused")
    for _ in range(5):
        tool(name="real-job")

    # it recovers, and two half-open probes succeed
    time.sleep(0.25)
    client.get_job.side_effect = None
    client.get_job.return_value = MagicMock(status="Running", runtime=None)
    tool(name="real-job")
    tool(name="real-job")

    # the agent mistypes a job name once
    client.get_job.side_effect = ApiException(status=404, reason="Not Found")
    tool(name="typo-job")

    # from here on the tool is dead, however long you wait
    client.get_job.side_effect = None
    for wait in (0.5, 1.0, 2.0):
        time.sleep(wait)
        print(tool(name="real-job")["error_code"])
```

```
CIRCUIT_OPEN
CIRCUIT_OPEN
CIRCUIT_OPEN
```

The breaker stays `HALF_OPEN` with all slots spent. Only a restart clears it.

## Proposed Fix

1. In `_audit_wrap`, record a success for any non-infrastructure result. The call completed; it just did not reach the Kubernetes API. This also makes the two call sites agree.
2. In `CircuitBreaker.can_execute()`, reopen a probe window that has been exhausted for longer than `recovery_timeout`, so a caller that fails to report cannot wedge the breaker.
3. Use `time.monotonic()` for every elapsed-time check in `resilience.py`.

## Acceptance Criteria

- A validation or not-found result resolves its half-open probe instead of consuming it silently.
- `full` and `progressive` / `semantic` modes account for the same tool result the same way.
- A breaker whose probes never report recovers on its own rather than staying closed to traffic.
- `half_open_max_calls` still caps concurrent probes.
- Wall-clock jumps no longer release the circuit early or wedge the rate limiter.

<br/>

**P.S.**: I've a working fix for this, and I'll shortly open a PR addressing the same. Thanks!

Neilblaze:
## Description

`can_execute()` reserves a half-open probe slot and only ever releases it when the probe reports an outcome. `_audit_wrap` reported nothing for a non-infrastructure result, so a single mistyped resource name during recovery could spend the last slot without a verdict. After that `can_execute()` refuses every call, nothing runs, nothing reports, and the breaker sits in `HALF_OPEN` until the process restarts — still telling callers it "retries automatically".

Three changes:

- `_audit_wrap` records a success only for a real success, and a failure for an infrastructure error. Anything else calls a new `CircuitBreaker.release()`, which hands a half-open slot back without counting as either, so a validation error can't reset the failure count during an outage.
- Each admission carries the probe generation, which moves on each time a half-open window opens, and reports from an older generation are ignored. Every admitted call now reports back, so there is no time-based reset, and a slow probe keeps its window.
- `resilience.py` measures elapsed time with `time.monotonic()`. `_audit_wrap` already did. A forward clock step used to hold the circuit open past `recovery_timeout`, and a backward one drove the rate limiter's bucket negative, so every tool returned `RATE_LIMITED` until it refilled.

The wedge test reproduces #221 on `main`, where the breaker stays `HALF_OPEN`, and the generation and release tests fail there too. `test_half_open_max_calls_limit` clears the `TODO(test)` marker that was sitting on that line.

No tool schema, parameter, or response shape changed.
ㅤ

## Related Issue
Fixes #221

## Checklist

## Testing
- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

> [!NOTE]
> `571 passed, 8 skipped`; `ruff check` / `ruff format --check` / `pre-commit run --all-files`
clean; tool schema snapshot unchanged.

Ran the reproduction from the issue before and after. Before, the tool returned `CIRCUIT_OPEN` forever. After, the mistyped name returns `RESOURCE_NOT_FOUND`, the next valid call succeeds, and the breaker closes.

### DIFF
--- a/kubeflow_mcp/core/resilience.py
+++ b/kubeflow_mcp/core/resilience.py
@@ -50,10 +50,18 @@ class CircuitBreaker:
     last_failure_time: float = field(default=0.0)
     half_open_calls: int = field(default=0)
     _half_open_successes: int = field(default=0)
+    _half_open_started: float = field(default=0.0)
     _lock: threading.Lock = field(init=False, repr=False)
 
     def __post_init__(self) -> None:
         self._lock = threading.Lock()
+
+    def _open_probe_window(self, now: float) -> None:
+        """Start a fresh batch of half-open probes. Caller holds the lock."""
+        self.state = CircuitState.HALF_OPEN
+        self.half_open_calls = 0
+        self._half_open_successes = 0
+        self._half_open_started = now
 
     def can_execute(self) -> bool:
         """Check if call is allowed and atomically reserve a slot if half-open."""
@@ -61,14 +69,25 @@ class CircuitBreaker:
             if self.state == CircuitState.CLOSED:
                 return True
 
+            now = time.monotonic()
+
             if self.state == CircuitState.OPEN:
-                if time.time() - self.last_failure_time >= self.recovery_timeout:
-                    self.state = CircuitState.HALF_OPEN
-                    self.half_open_calls = 0
+                if now - self.last_failure_time >= self.recovery_timeout:
+                    self._open_probe_window(now)
                     logger.info("Circuit breaker: OPEN -> HALF_OPEN")
                     self.half_open_calls += 1
                     return True
                 return False
+
+            # A probe that never reports back holds its slot for good, and the
+            # breaker only leaves HALF_OPEN once a probe does report. Reopening
+            # a stale window keeps that from becoming a permanent outage.
+            if (
+                self.half_open_calls >= self.half_open_max_calls
+                and now - self._half_open_started >= self.recovery_timeout
+            ):
+                self._open_probe_window(now)
+                logger.info("Circuit breaker: reopened a stale half-open probe window")
 
             if self.half_open_calls < self.half_open_max_calls:
                 self.half_open_calls += 1
@@ -92,7 +111,7 @@ class CircuitBreaker:
         """Record failed call."""
         with self._lock:
             self.failure_count += 1
-            self.last_failure_time = time.time()
+            self.last_failure_time = time.monotonic()
 
             if self.state == CircuitState.HALF_OPEN:
                 self.state = CircuitState.OPEN
@@ -244,13 +263,13 @@ class RateLimiter:
 
     def __post_init__(self) -> None:
         self._tokens = self.capacity
-        self._last_update = time.time()
+        self._last_update = time.monotonic()
         self._lock = threading.Lock()
 
     def acquire(self, tokens: float = 1.0) -> bool:
         """Try to acquire tokens. Returns True if successful."""
         with self._lock:
-            now = time.time()
+            now = time.monotonic()
             elapsed = now - self._last_update
             self._tokens = min(self.capacity, self._tokens + elapsed * self.rate)
             self._last_update = now
@@ -270,10 +289,10 @@ class SessionManager:
 
     def record_activity(self) -> None:
         """Record session activity."""
-        self._timestamps.append(time.time())
+        self._timestamps.append(time.monotonic())
 
     def is_stale(self) -> bool:
         """Check if session appears stale."""
         if not self._timestamps:
             return False
-        return time.time() - self._timestamps[-1] > self.max_age
+        return time.monotonic() - self._timestamps[-1] > self.max_age
--- a/kubeflow_mcp/core/resilience_test.py
+++ b/kubeflow_mcp/core/resilience_test.py
@@ -91,7 +91,43 @@ class TestCircuitBreaker:
         cb.record_failure()
         assert cb.state == CircuitState.OPEN
 
-    # TODO(test): test half_open_max_calls limit
+    @pytest.mark.slow
+    def test_half_open_max_calls_limit(self):
+        cb = CircuitBreaker(failure_threshold=1, recovery_timeout=0.05, half_open_max_calls=2)
+        cb.record_failure()
+        time.sleep(0.06)
+
+        assert cb.can_execute() is True
+        assert cb.can_execute() is True
+        assert cb.can_execute() is False
+        assert cb.half_open_calls == 2
+
+    @pytest.mark.slow
+    def test_half_open_reopens_when_a_probe_never_reports(self):
+        """A probe that records neither outcome must not wedge the breaker."""
+        cb = CircuitBreaker(failure_threshold=1, recovery_timeout=0.01, half_open_max_calls=2)
+        cb.record_failure()
+        time.sleep(0.02)
+
+        cb.can_execute()
+        cb.record_success()
+        cb.can_execute()  # this probe never reports back
+
+        assert cb.can_execute() is False
+        time.sleep(0.02)
+        assert cb.can_execute() is True
+        assert cb.state == CircuitState.HALF_OPEN
+
+    def test_recovery_ignores_wall_clock_jumps(self, monkeypatch):
+        """A wall-clock jump must not release the circuit before its timeout."""
+        cb = CircuitBreaker(failure_threshold=1, recovery_timeout=60.0)
+        cb.record_failure()
+
+        jumped = time.time() + 3600
+        monkeypatch.setattr("time.time", lambda: jumped)
+        assert cb.can_execute() is False
+        assert cb.state == CircuitState.OPEN
+
     # TODO(test): test thread safety with concurrent record_failure/record_success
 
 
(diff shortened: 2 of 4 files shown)


kubeflow_mcp/core/resilience_test.py
```diff
@@ -186,6 +222,12 @@ def test_tokens_refill_over_time(self):
         time.sleep(0.01)
         assert rl.acquire() is True
 
+    def test_acquire_ignores_wall_clock_jumps(self, monkeypatch):
```
abhijeet-dhumal, line 225: nit: this test patches time.time but limiter uses monotonic now.. can you consider patching monotonic for accuracy?

kubeflow_mcp/core/server.py
```diff
@@ -181,10 +181,14 @@ def wrapper(**kwargs):
                 )
                 span.set_attribute("tool.success", is_success)
                 span.set_attribute("tool.duration_ms", duration_ms)
-                if is_success:
-                    breaker.record_success()
-                elif is_infrastructure_error(result):
+                # A validation or not-found result still completed without an
```
abhijeet-dhumal: these comments seems unnecessary and can be made concise.. wdyt?

kubeflow_mcp/core/resilience_test.py
Neilblaze: @abhijeet-dhumal Thanks for the review!

I believe patching `time.time` is deliberate, & that's the clock the old code read, so it's what makes this fail on main. I did try monotonic and it comes out backwards, it passes on main and fails here. Stepping monotonic back drives elapsed negative, and the fix doesn't clamp, it just moves to a clock that can't step back.

So I added a docstring so it isn't a guessing game next time.

To your next, -- patching monotonic forward would be useful though, it'd let `test_tokens_refill_over_time` drop its sleep and slow marker. If you want, I can create a follow up once this gets merged? (if that sounds good?)

kubeflow_mcp/core/server.py
Neilblaze: Fair thing, I've addressed this in the new commit https://github.com/kubeflow/mcp-server/pull/222/commits/0ba13c32bc740152193fda5a260a2a17eb815424 :)

kubeflow_mcp/core/resilience.py
```diff
@@ -50,26 +50,45 @@ class CircuitBreaker:
...
                 return False
 
+            # A probe that never reports back holds its slot for good, and the
+            # breaker only leaves HALF_OPEN once a probe does report. Reopening
+            # a stale window keeps that from becoming a permanent outage.
+            if (
+                self.half_open_calls >= self.half_open_max_calls
+                and now - self._half_open_started >= self.recovery_timeout
+            ):
+                self._open_probe_window(now)
```
abhijeet-dhumal: Resetting the counters can let late probes from the previous window affect the new one. Please track probe generations or prevent overlapping windows.

kubeflow_mcp/core/server.py
```diff
@@ -181,10 +181,12 @@ def wrapper(**kwargs):
...
                 span.set_attribute("tool.duration_ms", duration_ms)
-                if is_success:
-                    breaker.record_success()
-                elif is_infrastructure_error(result):
+                # A non-infrastructure result still resolves the half-open probe,
+                # so it records a success here, matching execute_tool.
+                if is_infrastructure_error(result):
                     breaker.record_failure()
+                else:
+                    breaker.record_success()
```
abhijeet-dhumal: Validation/permission errors call record_success() and reset infrastructure failure counts. Only infrastructure successes should reset the closed-state breaker.

Neilblaze:
@abhijeet-dhumal thanks for the feedback, I've addressed those in the most recent commit. PTAL! :)

Neilblaze:
@abhijeet-dhumal PTAL when you get some time :)

abhishekKokadwar:
I pulled this branch to try it out, because the `time.time()` to `time.monotonic()` change is something I had independently run into from a different direction. The change looks right to me and `make verify` is clean. One thing to flag though: `test_half_open_max_calls_limit` is flaky on this branch, and it is a side effect of the monotonic switch rather than a pre-existing problem.

It failed 1 of 6 full-suite runs for me, then passed when run alone. Tightened into a loop it is about 10%:

```
Reproducing the failing sequence 400 times:
   run 3:  can_execute -> False, False, False; half_open_calls=0, state=open
   run 11: can_execute -> False, False, False; half_open_calls=0, state=open
   39/400 runs violated the assertion
```

The failure is the breaker never leaving OPEN, so the first `can_execute()` returns False instead of True.

**Cause.** The test uses `recovery_timeout=0.05` with `time.sleep(0.06)`, a 10 ms margin. `time.monotonic()` on Windows has a resolution of 15.625 ms, which is coarser than that margin, so a sleep that really did take 60 ms can measure as less than the 50 ms timeout:

```
monotonic resolution:    0.015625 s
perf_counter resolution: 1e-07 s

measured elapsed via monotonic after sleep(0.06):
   0.0620 s
   0.0630 s
   0.0470 s   <- below 0.05, breaker stays OPEN
   0.0620 s
```

`time.time()` on Windows is backed by a higher-resolution source, so this margin happened to hold before. The switch to monotonic is still the correct change, the test margin just was not sized for it. `test_full_window_is_not_replaced_while_probes_are_out` has the same shape (`recovery_timeout=0.01`, `sleep(0.02)`) and looks exposed for the same reason, though I did not catch it failing.

**Two options.** Widening the sleep works (0/200 with `sleep(0.2)`) but slows the suite and only makes the race rarer. Driving the clock removes it:

```python
with patch("kubeflow_mcp.core.resilience.time.monotonic", lambda: clock["t"]):
    cb = CircuitBreaker(failure_threshold=1, recovery_timeout=0.05, half_open_max_calls=2)
    cb.record_failure()
    clock["t"] += 0.06
    assert (cb.can_execute(), cb.can_execute(), cb.can_execute()) == (True, True, False)
    assert cb.half_open_calls == 2
```

That is stable across runs, needs no sleep, and drops the `@pytest.mark.slow` marker. It is also the same technique your `test_rate_limiter_*` tests already use with `monkeypatch.setattr("time.time", ...)`, so it would be consistent with the rest of the file. Those two will need retargeting to `time.monotonic` on this branch as well, since that is the clock the limiter reads now.

For what it is worth, the five behaviours this PR fixes all reproduce for me on `main`, and are worth keeping in the description if it helps reviewers: breaker stuck after a backward step, breaker opening early after a forward step, rate limiter refilling instantly on a forward step, rate limiter stalling on a backward step, and a fresh session reported stale. Not a maintainer, so treat this as a suggestion.

PR #224, opened 2026-09-07
fix(core): redact credentials consistently across log paths

### LINKED ISSUE #223: fix(core): bearer credentials survive log redaction
> [!IMPORTANT]
> I'm aware that the mcp-server is on v0.1.1 and the [ROADMAP](https://github.com/kubeflow/mcp-server/blob/main/ROADMAP.md) still marks it as Development, so I am not filing this as a production incident. I am opening it as a hardening fix, because the redaction code is already there and looks like it was meant to cover this case, and lining the paths up now seems cheaper than doing it once more code leans on them.

ㅤ
## Problem

`kubeflow_mcp/core/logging.py` redacts free text in three different places, and the three do not agree with each other.

`BufferingHandler.emit()` redacts the message with its own substitution. That buffer is what `get_server_logs()` hands back, and the tool sits in the `readonly` persona, so whatever survives goes to the model.

`StructuredFormatter.format()` writes `record.getMessage()` and the formatted exception straight through, with no redaction at all. `setup_logging()` picks this formatter whenever stderr is not a TTY, which is what a container gets, so that text ends up in the log collector.

`_apply_pattern()` redacts string leaves of structured extras by replacing the whole match.

All three share one regex, and it misses the most common shape a credential turns up in.

```python
_REDACT_PATTERNS = re.compile(
    r"(token|password|secret|bearer|authorization|credential)[=: ]+\S+",
    re.IGNORECASE,
)
```

Given `Authorization: Bearer <jwt>`, that single `\S+` gets eaten by the scheme word `Bearer`, so the match stops short and the JWT stays in the line.

Worth flagging that `_audit_wrap` logs its failure path with `exc_info=True`, so any exception carrying a token in its message reaches stderr in full.
ㅤ

## Reproduction

```python
import json, logging
from kubeflow_mcp.core.logging import BufferingHandler, StructuredFormatter, get_log_buffer

JWT = "eyJhbGciOiJIUzI1NiJ9.SECRETPAYLOAD.sig"
line = f"calling API with Authorization: Bearer {JWT}"
record = logging.LogRecord("kubeflow_mcp.test", logging.WARNING, __file__, 1, line, (), None)

BufferingHandler().emit(record)
print("buffer", get_log_buffer()[-1]["message"])
print("stderr", json.loads(StructuredFormatter().format(record))["message"])
```

```
buffer calling API with Authorization: *** eyJhbGciOiJIUzI1NiJ9.SECRETPAYLOAD.sig
stderr calling API with Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.SECRETPAYLOAD.sig
```

The buffer redacts the word `Bearer` and keeps the JWT. stderr keeps the whole line. The exception text on the `tool_call_failed` path behaves the same way.
ㅤ

## Proposed Fix

- Give the regex an optional auth scheme group, so the credential that follows `Bearer`, `Basic`
or `Digest` is part of the match.
- Route every path through a single `_redact_text()` helper. It can keep the whole match
substitution `_apply_pattern` already uses, so a line reads the same wherever it lands.
- Redact the message and the formatted exception in `StructuredFormatter`, and the message in
`ConsoleFormatter`.

## Acceptance Criteria

- A credential following an auth scheme is removed, not just the scheme word.
- The buffer, the JSON formatter and the console formatter produce the same redacted message.
- Exception text on the failure path is redacted before it reaches stderr.
- Ordinary log lines stay untouched, including parameter names like `image_pull_secrets`.

<br/>

**P.S.**: I'll shortly open a PR addressing this. Thanks!

Neilblaze:
## Description

`core/logging.py` handled free text in three places, with two different substitutions and one path that did nothing at all.

`BufferingHandler.emit()` used a custom lambda. Its output feeds `get_server_logs()`, which the `readonly` persona can call, so it goes back to the model.

`StructuredFormatter.format()` wrote the message and the formatted exception verbatim. It is the formatter chosen whenever stderr is not a TTY, so this is what a container ships to its log collector.

`_apply_pattern()` used a whole match substitution on structured extras.

The shared regex also stopped one token too early. In `Authorization: Bearer <jwt>` the `\S+` was taken by `Bearer`, so the credential itself never matched. The buffer redacted the scheme word and kept the JWT, and stderr kept the whole line.

### What changed 

- The value match takes an optional auth scheme group, so `Bearer`, `Basic`, `Digest` or `Token` followed by a credential is matched in full.
- One `_redact_text()` helper, used by the buffer, both formatters and `_apply_pattern`. It keeps the whole match substitution `_apply_pattern` already used, the one that `test_redacts_pattern_inside_non_sensitive_key_value` pins, so structured redaction is unchanged.
- Keys are judged by `is_sensitive_key()`, pulled out of `mask_sensitive_data` in `core/security.py`, so free text and structured fields share one vocabulary. Compound names like `secret_access_key`, `private_key` and `api_key` are caught, and the safe keys stay untouched.
- The shared `_token` suffix rule is now `token`, so glued names like `accessToken` keep being caught in both paths.
- `StructuredFormatter` redacts the message and the formatted exception, and `ConsoleFormatter` redacts the message.

I checked for false positives with an AST walk over the 54 statically reconstructable `logger.*` messages in the package, and none of them changes. `image_pull_secrets not set` and the `cli.py` auth warning are both left alone, because neither has a key followed by a real `=` or `:`.
ㅤ

## Related Issue
Fixes #223
ㅤ

## Checklist
ㅤ

## Testing
- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Manually tested (described below)

> [!NOTE]
> `586 passed, 8 skipped`. `ruff check`, `ruff format --check` and `pre-commit run --all-files` are clean, and the tool schema snapshot is unchanged.

I ran the reproduction from the issue against `main` and against this branch. On `main` the JWT survives in the buffer, on stderr and in the exception field, and the buffer and stderr disagree with each other. On this branch all three drop it, and the buffer and stderr messages come out identical.

New coverage is `TestRedactText` for the helper itself, and `TestLogPathsRedactConsistently` for the buffer, the JSON formatter, the console formatter, the exception text, plus a case asserting the three paths agree on the same message.

### DIFF
--- a/kubeflow_mcp/core/logging.py
+++ b/kubeflow_mcp/core/logging.py
@@ -31,6 +31,26 @@ request_context: ContextVar[dict[str, Any] | None] = ContextVar("request_context
 
 _log_buffer: deque[dict[str, Any]] = deque(maxlen=1000)
 
+# The credential may be preceded by an auth scheme, as in
+# ``Authorization: Bearer <jwt>``. Without the optional scheme group the match
+# ends at ``Bearer`` and leaves the credential itself in the line.
+_REDACT_PATTERNS = re.compile(
+    r"(token|password|secret|bearer|authorization|credential)"
+    r"[=: ]+"
+    r"(?:(?:bearer|basic|digest|token)\s+)?"
+    r"\S+",
+    re.IGNORECASE,
+)
+
+
+def _redact_text(text: str) -> str:
+    """Redact credential values in free text.
+
+    Every log path routes through here so a line is redacted the same way
+    whether it lands in the buffer, on stderr, or inside a structured field.
+    """
+    return _REDACT_PATTERNS.sub("***", text)
+
 
 def _redact_dict(d: Any) -> Any:
     """Recursively redact sensitive data for logging."""
@@ -50,7 +70,7 @@ def _apply_pattern(d: Any) -> Any:
     if isinstance(d, list):
         return [_apply_pattern(v) for v in d]
     if isinstance(d, str):
-        return _REDACT_PATTERNS.sub("***", d)
+        return _redact_text(d)
 
     return d
 
@@ -63,12 +83,12 @@ class StructuredFormatter(logging.Formatter):
             "timestamp": datetime.now(timezone.utc).isoformat(),
             "level": record.levelname,
             "logger": record.name,
-            "message": record.getMessage(),
+            "message": _redact_text(record.getMessage()),
             "correlation_id": correlation_id.get() or None,
         }
 
         if record.exc_info:
-            log_dict["exception"] = self.formatException(record.exc_info)
+            log_dict["exception"] = _redact_text(self.formatException(record.exc_info))
 
         ctx = request_context.get()
         if ctx is not None:
@@ -100,36 +120,20 @@ class ConsoleFormatter(logging.Formatter):
         cid = correlation_id.get()
         cid_str = f" [{cid[:8]}]" if cid else ""
         return (
-            f"{color}{record.levelname:8}{self.RESET}{cid_str} {record.name}: {record.getMessage()}"
+            f"{color}{record.levelname:8}{self.RESET}{cid_str} "
+            f"{record.name}: {_redact_text(record.getMessage())}"
         )
-
-
-_REDACT_PATTERNS = re.compile(
-    r"(token|password|secret|bearer|authorization|credential)[=: ]+\S+",
-    re.IGNORECASE,
-)
 
 
 class BufferingHandler(logging.Handler):
     """Handler that stores logs in memory buffer with sensitive data redacted."""
 
     def emit(self, record: logging.LogRecord) -> None:
-        message = record.getMessage()
-        message = _REDACT_PATTERNS.sub(
-            lambda m: (
-                m.group().split("=")[0] + "=***"
-                if "=" in m.group()
-                else m.group().split(":")[0] + ": ***"
-                if ":" in m.group()
-                else m.group().split()[0] + " ***"
-            ),
-            message,
-        )
         log_entry = {
             "timestamp": datetime.now(timezone.utc).isoformat(),
             "level": record.levelname,
             "logger": record.name,
-            "message": message,
+            "message": _redact_text(record.getMessage()),
         }
         _log_buffer.append(log_entry)
 
--- a/kubeflow_mcp/core/logging.py
+++ b/kubeflow_mcp/core/logging.py
@@ -31,12 +31,12 @@ request_context: ContextVar[dict[str, Any] | None] = ContextVar("request_context
 
 _log_buffer: deque[dict[str, Any]] = deque(maxlen=1000)
 
-# The credential may be preceded by an auth scheme, as in
-# ``Authorization: Bearer <jwt>``. Without the optional scheme group the match
-# ends at ``Bearer`` and leaves the credential itself in the line.
+# The scheme word is matched too, as in ``Authorization: Bearer <jwt>``. The
+# separator has to be a real ``=`` or ``:``, otherwise prose like ``bearer auth``
+# matches and ordinary log lines get mangled.
 _REDACT_PATTERNS = re.compile(
     r"(token|password|secret|bearer|authorization|credential)"
-    r"[=: ]+"
+    r"\s*[=:]\s*"
     r"(?:(?:bearer|basic|digest|token)\s+)?"
     r"\S+",
     re.IGNORECASE,
--- a/kubeflow_mcp/core/logging_test.py
+++ b/kubeflow_mcp/core/logging_test.py
@@ -16,10 +16,21 @@
 
 import json
 import logging
+import sys
 
 import pytest
 
-from kubeflow_mcp.core.logging import StructuredFormatter, _redact_dict, request_context
+from kubeflow_mcp.core.logging import (
+    BufferingHandler,
+    ConsoleFormatter,
+    StructuredFormatter,
+    _redact_dict,
+    _redact_text,
+    get_log_buffer,
+    request_context,
+)
+
+_JWT = "eyJhbGciOiJIUzI1NiJ9.SECRETPAYLOAD.sig"
 
 
 class TestRedactDict:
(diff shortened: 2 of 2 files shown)


kubeflow_mcp/core/logging.py
```diff
@@ -31,6 +31,26 @@
...
 _log_buffer: deque[dict[str, Any]] = deque(maxlen=1000)
 
+# The credential may be preceded by an auth scheme, as in
+# ``Authorization: Bearer <jwt>``. Without the optional scheme group the match
+# ends at ``Bearer`` and leaves the credential itself in the line.
+_REDACT_PATTERNS = re.compile(
+    r"(token|password|secret|bearer|authorization|credential)"
+    r"[=: ]+"
+    r"(?:(?:bearer|basic|digest|token)\s+)?"
+    r"\S+",
```
abhijeet-dhumal, line 41: On main branch, \S+ stops at Bearer and the JWT survives .. is that the failure mode you saw reproducing #223? Does the optional scheme group cover any other auth header shapes we should parametrize (e.g. Token <value>)?

kubeflow_mcp/core/logging.py
```diff
@@ -31,6 +31,26 @@
...
+)
+
+
+def _redact_text(text: str) -> str:
+    """Redact credential values in free text.
+
+    Every log path routes through here so a line is redacted the same way
+    whether it lands in the buffer, on stderr, or inside a structured field.
+    """
+    return _REDACT_PATTERNS.sub("***", text)
```
abhijeet-dhumal, line 52: Are there any log paths left that still bypass _redact_text()?

kubeflow_mcp/core/logging.py
```diff
@@ -63,12 +83,12 @@ def format(self, record: logging.LogRecord) -> str:
             "timestamp": datetime.now(timezone.utc).isoformat(),
             "level": record.levelname,
             "logger": record.name,
-            "message": record.getMessage(),
+            "message": _redact_text(record.getMessage()),
```
abhijeet-dhumal, line 86: is this the path that was leaking credentials to the log collector?

kubeflow_mcp/core/logging.py
```diff
@@ -100,36 +120,20 @@ def format(self, record: logging.LogRecord) -> str:
...
-_REDACT_PATTERNS = re.compile(
-    r"(token|password|secret|bearer|authorization|credential)[=: ]+\S+",
-    re.IGNORECASE,
-)
-
-
 class BufferingHandler(logging.Handler):
     """Handler that stores logs in memory buffer with sensitive data redacted."""
 
     def emit(self, record: logging.LogRecord) -> None:
```
abhijeet-dhumal, line 131: The old buffer lambda kept a prefix like `Authorization: Bearer ***` but left the JWT .. did you consider whether callers relied on that partial format, or is full-match `***` the behaviour we want everywhere now?

kubeflow_mcp/core/logging_test.py
```diff
@@ -163,3 +174,76 @@ def test_no_context_omits_context_key(self):
...
+        result = _redact_text(text)
+        assert leaked not in result
+        assert "***" in result
+
+    def test_plain_message_untouched(self):
+        assert _redact_text("tool_call completed in 12ms") == "tool_call completed in 12ms"
+
+    def test_word_ending_in_keyword_not_matched(self):
+        # image_pull_secrets is a parameter name, not a credential assignment.
+        assert _redact_text("image_pull_secrets not set") == "image_pull_secrets not set"
```
abhijeet-dhumal, line 202: You guarded image_pull_secrets here  did you hit any false positives scanning the ~39 `logger.*` call sites? e.g. does the cli.py auth warning (`Set --auth-token or KUBEFLOW_MCP_AUTH_TOKEN...`) come out readable after redaction?

kubeflow_mcp/core/logging_test.py
```diff
@@ -163,3 +174,76 @@ def test_no_context_omits_context_key(self):
...
+    def test_structured_formatter_redacts_exception_text(self):
+        try:
+            raise RuntimeError(f"boom: token={_JWT}")
+        except RuntimeError:
+            record = self._record("tool_call_failed", exc_info=sys.exc_info())
+
+        output = json.loads(StructuredFormatter().format(record))
+        assert _JWT not in output["exception"]
+
+    def test_all_paths_produce_the_same_message(self):
```
abhijeet-dhumal, line 249: This test seems the key regression test .. should we also assert the redacted output is identical (not just that the JWT is absent), to catch future drift between paths?

abhijeet-dhumal:
Thanks @Neilblaze, this looks like the right fix for #223.  I have left few questions before lgtm ⬆️

kubeflow_mcp/core/logging.py
Neilblaze: @abhijeet-dhumal Thanks for the review!

Yes, that's exactly the failure mode in #223. On main the `\S+` gets eaten by `Bearer`, so the match ends there & the JWT stays in the line.

`Token <value>` is already covered, the optional group is `bearer|basic|digest|token`. I checked `Token` and `Digest` and both come out redacted. Happy to parametrize the test with those two if you want them pinned.

kubeflow_mcp/core/logging.py
Neilblaze: One, string valued extras. `StructuredFormatter` only runs `_redact_dict` when the extra is a dict or list, so a string one goes through untouched. Right now the only string extra in the allowlist is `tool`, which is the tool function name, so nothing leaks in practice. Easy to tighten here if you'd rather not leave the hole open.

Also worth saying `setup_logging` only wires up the `kubeflow_mcp` logger, so third party loggers like the k8s client never reach our handlers at all -- that felt like a separate job.

kubeflow_mcp/core/logging.py
Neilblaze: Yes, that's the one. `setup_logging` picks the JSON formatter whenever stderr isn't a TTY, so a container always lands here, and the message went out raw.

kubeflow_mcp/core/logging.py
Neilblaze: Close, though the old lambda actually gave `Authorization: *** <jwt>`, so it redacted the scheme word and kept the credential. Nothing depended on that shape, upstream `logging_test.py` never touched `BufferingHandler` or `get_log_buffer`.

Full match `***` is what I went with because `_apply_pattern` already did that and `test_redacts_pattern_inside_non_sensitive_key_value` pins it. That way there's one behaviour and no existing test had to move.

kubeflow_mcp/core/logging_test.py
Neilblaze: Good catch, and you picked the exact one. 

My earlier check was a grep that only saw the first string literal per call, so it missed messages built from concatenated lines. Redid it as an AST walk over 54 messages and that cli.py warning is the single hit. It comes out as `Set --auth-*** KUBEFLOW_MCP_AUTH_*** *** or ...`, so three mangles in one line.

Turns out, the weakness is in the pattern itself, `[=: ]+` lets a bare space count as a separator, so `token or` and `bearer auth,` match. It's on main too, it just never showed because the JSON formatter didn't redact messages back then, so this PR is what surfaces it.

Given the case, the fix is to require an actual `=` or `:`, so `\s*[=:]\s*`. Every real shape still redacts, including `token = spaced value`, and the rescan comes back clean. Pushing that with a test for the warning so it can't drift back.

kubeflow_mcp/core/logging_test.py
Neilblaze: It does already, `test_all_paths_produce_the_same_message` ends with `assert buffered == structured`. The hunk cut off right at the def line so it's easy to miss.

As you said, I can go further and pin the exact expected string rather than just equality between the paths, which would also catch drift in the substitution itself. I'll push that in.

kubeflow_mcp/core/logging.py
```diff
@@ -31,6 +31,26 @@
 
 _log_buffer: deque[dict[str, Any]] = deque(maxlen=1000)
 
+# The scheme word is matched too, as in ``Authorization: Bearer <jwt>``. The
+# separator has to be a real ``=`` or ``:``, otherwise prose like ``bearer auth``
+# matches and ordinary log lines get mangled.
+_REDACT_PATTERNS = re.compile(
+    r"(token|password|secret|bearer|authorization|credential)"
```
abhijeet-dhumal: This pattern does not match common sensitive keys such as secret_access_key=..., private_key=..., or api_key=... because it only recognizes the standalone keywords before = or : 
Since _redact_text now covers exception and message paths, these values can still be logged unredacted.

abhijeet-dhumal: Please include the compound credential names (or reuse the existing sensitive-key vocabulary) and add regression tests while preserving safe-key exceptions.

Neilblaze: Thanks, good catch. I've addressed it appropriately in the most recent commit, PTAL! :)

Neilblaze:
@abhijeet-dhumal I think we're good to go! :)

abhijeet-dhumal:
Thanks @Neilblaze !
/lgtm
/approve

PR #230, opened 2026-09-12
fix(core): handle non-ASCII tokens in API key verification

### LINKED ISSUE #229: Non-ASCII bearer token returns 500 instead of 401
### Description

`APIKeyVerifier.verify_token` compares two `str` values with `hmac.compare_digest` (`kubeflow_mcp/core/auth.py` L64). Python raises `TypeError` when either string has non-ASCII characters, so the request fails with 500 instead of being rejected.

### Steps to Reproduce

1. `kubeflow-mcp serve --transport http --auth-token s3cret`
2. POST an `initialize` request to `/mcp` with header `Authorization: Bearer café` (UTF-8)
3. See 500

### Expected Behavior

401 with `invalid_token`, same as a wrong ASCII token.

### Actual Behavior

500 Internal Server Error. Log ends with:

```
TypeError: comparing strings with non-ASCII characters is not supported
```

(from `auth.py` line 64, in `verify_token`)

### MCP Server Version

0.1.1 (main at 0f90ca6)

### Python Version

3.12

### Kubernetes Version

n/a

### MCP Client

Custom / Other

### References

MCP authorization spec, invalid or expired tokens get HTTP 401. Fix is to compare bytes, `hmac.compare_digest(token.encode(), self._expected.encode())`, plus a test in `auth_test.py`.

krishna231122-del:
## Description

Fixes a `TypeError` in `APIKeyVerifier.verify_token` that occurs when checking non-ASCII tokens. The `hmac.compare_digest` function only supports ASCII characters when passing strings. This PR updates the comparison to explicitly encode both the incoming token and the expected token to bytes via `.encode()`. This ensures that invalid tokens containing non-ASCII characters (e.g., `café`) are gracefully rejected with a 401 Unauthorized response instead of crashing the server with a 500 Internal Server Error.

A test case `test_rejects_non_ascii_token` has been added to `auth_test.py` to prevent future regressions.

## Related Issue
Fixes #229 

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

**Manual Testing Details:**
Confirmed that passing an Authorization header with a non-ASCII string to the server no longer yields a 500 status code and correctly returns a 401 instead. Unit tests were run using `uv run pytest kubeflow_mcp/core/auth_test.py` to ensure the logic behaves as expected.


abhijeet-dhumal:
Thanks @krishna231122-del!
The code change looks good and I found no issues. Please resolve the failing DCO check
once it passes, this is lgtm.

krishna231122-del:
Thanks for the review! Just fixed the missing sign-off for the DCO check, so everything should be green now. ✅

abhijeet-dhumal:
Thanks @krishna231122-del !
/lgtm
approve

abhijeet-dhumal:
/hold
@krishna231122-del The implementation looks correct, but the current title/commit subject `bug` does not follow the repository’s required Conventional Commit format. Please rename it, for example: `fix(core): handle non-ASCII API keys`.

kubeflow_mcp/core/auth_test.py
```diff
@@ -32,6 +32,11 @@ async def test_rejects_invalid_token(self):
         result = await verifier.verify_token("wrong-token")
         assert result is None
 
+    async def test_rejects_non_ascii_token(self):
+        verifier = APIKeyVerifier(expected_token="test-secret")
+        result = await verifier.verify_token("café")
+        assert result is None
```
abhijeet-dhumal, line 38: This test covers rejection of a non-ASCII token, but it does not verify that a valid configured non-ASCII token succeeds. Please add a case with expected_token="café" and token "café", plus a mismatched Unicode token case, to cover both sides of the new byte comparison.

krishna231122-del: I've added the requested test cases in auth_test.py to cover scenarios where the configured token itself is non-ASCII.

Specifically, I added:

test_accepts_valid_non_ascii_token: Verifies that a valid matching non-ASCII token (e.g., "café") succeeds.
test_rejects_mismatched_non_ascii_token: Verifies that a mismatched non-ASCII token correctly fails the byte comparison.
The tests pass successfully and I've pushed the updates to the branch. Let me know if everything looks good now!

PR #231, opened 2026-09-12
fix(core): reject non-ASCII bearer tokens with 401 instead of 500

### LINKED ISSUE #229: Non-ASCII bearer token returns 500 instead of 401

Ahmad-Faraj:
## Description

`APIKeyVerifier.verify_token` compared two `str` values with `hmac.compare_digest`, which raises `TypeError` when either has non-ASCII characters. A token like `café` crashed the request with 500. Comparing the UTF-8 bytes keeps the check constant-time and rejects it with 401 like any other wrong token.

## Related Issue

Fixes #229

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

Added `test_rejects_non_ascii_token` to `auth_test.py`. It fails on main with the `TypeError` and passes with this change.


Ahmad-Faraj:
Duplicate of #230.

PR #233, opened 2026-09-12
fix(trainer): accept dotted runtime names in get_runtime and list_training_jobs

### LINKED ISSUE #232: get_runtime rejects the torchtune runtime names Trainer ships
### Description

`get_runtime` and the `runtime` filter of `list_training_jobs` return `VALIDATION_ERROR` for any runtime name with a dot in it. All three torchtune runtimes that Kubeflow Trainer installs have one: `torchtune-llama3.2-1b`, `torchtune-llama3.2-3b` and `torchtune-qwen2.5-1.5b`. These are the runtimes `fine_tune()` needs, and its own error message suggests `torchtune-llama3.2-1b`.

ClusterTrainingRuntime names are Kubernetes object names, and dots are allowed there. The check came in with #163, so 0.1.1 is not affected.

### Steps to Reproduce

```python
from kubeflow_mcp.trainer.api.discovery import get_runtime, list_training_jobs
print(get_runtime("torchtune-llama3.2-1b"))
print(list_training_jobs(runtime="torchtune-qwen2.5-1.5b"))
```

No cluster needed. Against a cluster with Trainer v2.3.0 and its default runtimes, `list_runtimes` returns `torchtune-llama3.2-1b` and `get_runtime` on that same name fails.

### Expected Behavior

`get_runtime` returns the runtime and `list_training_jobs` filters by it.

### Actual Behavior

```
{'success': False, 'error': 'name must be lowercase alphanumeric with hyphens', 'error_code': 'VALIDATION_ERROR', 'details': {'value': 'torchtune-llama3.2-1b', 'pattern': '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'}, 'hint': None}
```

The `list_training_jobs` call fails the same way with `runtime must be lowercase alphanumeric with hyphens`.

### MCP Server Version

main at 0f90ca6

### Python Version

3.12

### Kubernetes Version

1.34

### MCP Client

Custom / Other

### References

#162, #163, https://github.com/kubeflow/trainer/tree/v2.3.0/manifests/base/runtimes/torchtune

I'll open a PR for this.

Ahmad-Faraj:
## Description

`get_runtime` and the `runtime` filter of `list_training_jobs` used the job name check, which allows only lowercase letters, digits and hyphens. Runtime names are Kubernetes object names and can have dots, so the three torchtune runtimes Trainer ships (`torchtune-llama3.2-1b`, `torchtune-llama3.2-3b`, `torchtune-qwen2.5-1.5b`) were rejected with `VALIDATION_ERROR`.

This adds `validate_runtime_name`, which follows the Kubernetes object name rules (dots allowed, up to 253 characters), and uses it for those two runtime checks. Job name checks are unchanged.

## Related Issue

Fixes #232

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

`test_validate_runtime_name` accepts the three stock names and rejects empty, uppercase, `..` and path traversal. `test_accepts_dotted_runtime_names` checks that both tools reach the SDK for dotted names. Both fail on main.


kubeflow_mcp/core/security.py
```diff
@@ -23,6 +23,10 @@
 
 K8S_NAME_PATTERN = re.compile(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$")
 MAX_NAME_LENGTH = 63
+K8S_SUBDOMAIN_PATTERN = re.compile(
+    r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
```
abhijeet-dhumal, line 27: Please enforce the Kubernetes 63-character limit for each dot-separated label. A 64-character label currently passes validation but fails later at the SDK/Kubernetes layer.. add a regression test.

Ahmad-Faraj: Done in `d860061`, with a test for a 64-character segment.

abhijeet-dhumal:
@Ahmad-Faraj can you resolve merge conflicts ? 
And please reuse the corrected validator from #236 and avoid maintaining a duplicate implementation

Ahmad-Faraj:
Resolved, and it now uses `validate_runtime_name` from #236 instead of its own copy.

Ahmad-Faraj:
@abhijeet-dhumal could you take another look when you have time?

PR #165, opened 2026-08-22
fix: validate monitoring job names

pratik-naik003:
## Description

Validates training job names at the start of all monitoring tools before calling the Trainer/Kubernetes SDK.

Updated:

- `get_training_logs()`
- `get_training_events()`
- `wait_for_training()`

The previous validation inside the `get_training_logs()` fallback path was removed because validation now happens once at the tool boundary.

Focused tests verify that invalid names return `VALIDATION_ERROR` and do not call the SDK.

## Related Issue

Fixes #164

## Testing

- [x] Focused monitoring tests
- [x] Full Python test suite (`221 passed`)
- [x] Ruff lint and formatting checks
- [x] Pre-commit checks

## Checklist

- [x] My commit is signed off
- [x] Tests pass locally
- [x] Linting and formatting pass

PR #163, opened 2026-08-22
fix: validate discovery resource names

pratik-naik003:
## Description

Validates Kubernetes resource names in Trainer discovery tools before SDK calls.

This adds validation for:

- `get_training_job(name=...)`
- `get_runtime(name=...)`
- `list_training_jobs(runtime=...)`

Invalid names now return the standard `VALIDATION_ERROR` response instead of reaching the Trainer/Kubernetes SDK and producing a less clear SDK error.

Focused unit tests also verify that the SDK client is not called when validation fails.

## Related Issue

Fixes #162

## Testing

- [x] Focused discovery validation tests
- [x] Full Python test suite (`221 passed`)
- [x] Ruff lint and formatting checks
- [x] Pre-commit checks

## Checklist

- [x] My commit is signed off
- [x] Tests pass locally
- [x] Linting and formatting pass

PR #236, opened 2026-09-13
fix(trainer): validate runtime names in platform runtime tools

### LINKED ISSUE #235: fix(trainer): platform runtime tools do not validate runtime names
### Description

The three runtime tools in `kubeflow_mcp/trainer/api/platform.py` (`create_runtime`, `patch_runtime` and `delete_runtime`) pass `name` straight to the Kubernetes API without validating it. `platform.py` does not import a name validator at all.

Occurrences of `validate_k8s_name` per trainer module on `main`:

| module | calls |
|---|---|
| `discovery.py` | 4 |
| `monitoring.py` | 4 |
| `training.py` | 4 |
| `lifecycle.py` | 3 |
| **`platform.py`** | **0** |

`patch_runtime` and `create_runtime` do validate their `patch` / `spec` dictionaries, so the omission looks like an oversight rather than a decision.

This is the same rollout #162 → #163 (discovery) and #164 → #165 (monitoring) completed for the other modules. #162 noted at the time that *"other Trainer APIs, such as `delete_training_job()` and `update_training_job()`, already use `validate_k8s_name()` before making SDK calls."* The platform runtime tools were never in scope of either PR.

### Steps to Reproduce

No cluster required:

```python
from unittest.mock import MagicMock, patch
from kubeflow_mcp.trainer.api.platform import delete_runtime

api = MagicMock()
api.list_cluster_custom_object.return_value = {"items": []}
api.delete_cluster_custom_object.return_value = {"status": "Success"}

with patch("kubeflow_mcp.common.utils.get_custom_objects_api", return_value=api):
    for name in ("", "   ", "UPPER-Case", "bad_underscore", "name with space", "../escape"):
        result = delete_runtime(name, confirmed=True)
        print(repr(name), "->", result.get("success"),
              "| api called:", api.delete_cluster_custom_object.called)
        api.delete_cluster_custom_object.reset_mock()
```

### Expected Behavior

`VALIDATION_ERROR` before the SDK/API call, matching `discovery.get_runtime` given the same input.

### Actual Behavior

On `main` at 0f90ca6, every one of those names is accepted, the API call is issued, and the tool reports success:

```
''                -> True | api called: True
'   '             -> True | api called: True
'UPPER-Case'      -> True | api called: True
'bad_underscore'  -> True | api called: True
'name with space' -> True | api called: True
'../escape'       -> True | api called: True
```

`patch_runtime("../escape", patch={...}, confirmed=True)` behaves the same. For contrast, all six names sent to `discovery.get_runtime` return `VALIDATION_ERROR`.

**Scope of the claim.** The Kubernetes client URL-encodes path segments and the API server would itself reject most of these names, so this is defence-in-depth and contract consistency rather than a demonstrated privilege escalation, I could not show one, and am not claiming it. The concrete defects are:

1. The three tools disagree with every sibling trainer module on identical input.
2. `success: True` is returned for names that cannot reference a real object.
3. An empty name reaches a cluster-scoped delete call.

These tools are `platform-admin` persona only, which limits exposure.

### Proposed fix

Validate `name` first in each of the three tools, matching the `discovery.py` pattern exactly (early return, no shared wrapper, since a refactor would be larger than the defect).

One wrinkle: `validate_k8s_name` is **too strict** for runtimes. Runtime names are Kubernetes object names and may contain dots. The torchtune runtimes Trainer ships are `torchtune-llama3.2-1b` and `torchtune-qwen2.5-1.5b`. Using `validate_k8s_name` here would reintroduce exactly the bug reported in #232. So this adds a subdomain-form validator:

```python
K8S_SUBDOMAIN_PATTERN = re.compile(
    r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
)
MAX_SUBDOMAIN_LENGTH = 253


def validate_runtime_name(name: str, field: str = "name") -> ToolError | None:
    """Runtime names are K8s object names, so unlike job names they may contain dots."""
```

**Relationship to #232 / #233.** #233 introduces a `validate_runtime_name` with the same name, pattern and limit for the discovery tools. I deliberately matched its signature and semantics so the two converge on one helper rather than conflicting, whichever lands first, the second becomes a small rebase. Happy to rebase onto #233, or to split the helper into its own PR first if you would rather land it once and have both build on it.

### MCP Server Version

0.1.1 (`main` at 0f90ca6)

### Python Version

3.11

### Kubernetes Version

n/a. Reproduces with a mocked CustomObjects API

### MCP Client

Custom / Other

### References

- #162 / #163: same validation rollout for discovery tools
- #164 / #165: same rollout for monitoring tools
- #232 / #233: dotted runtime names; why `validate_k8s_name` is not the right helper here
- #97: added unit tests for these runtime tools (validation was not part of that scope)

I have a branch ready with the fix and tests (33 new tests, `594 passed`, `make verify` clean) and will open a PR shortly.

abhishekKokadwar:
## Description

`create_runtime`, `patch_runtime` and `delete_runtime` in `kubeflow_mcp/trainer/api/platform.py` passed `name` straight to the Kubernetes API without validating it. `platform.py` did not import a name validator at all, while every other trainer module does:

| module | `validate_*_name` calls before | after |
|---|---|---|
| `discovery.py` | 4 | 4 |
| `monitoring.py` | 4 | 4 |
| `training.py` | 4 | 4 |
| `lifecycle.py` | 3 | 3 |
| **`platform.py`** | **0** | **3** |

#163 and #165 rolled this validation out to the discovery and monitoring tools; the platform runtime tools were never in scope of either. This completes that rollout for the three cluster-scoped mutating tools.

### What changes

1. **`core/security.py`**: adds `validate_runtime_name()` plus `K8S_SUBDOMAIN_PATTERN` / `MAX_SUBDOMAIN_LENGTH` (253).
2. **`trainer/api/platform.py`**: imports it and validates `name` first in the three runtime tools, matching the `discovery.py` early-return pattern.

No tool names, parameters, or response shapes change. The only behavioural change is that an invalid `name` now returns `VALIDATION_ERROR` instead of reporting success, which is what the other four modules already do.

### Why not `validate_k8s_name`

Runtime names are Kubernetes object names and may contain dots. The torchtune runtimes Trainer ships are `torchtune-llama3.2-1b` and `torchtune-qwen2.5-1.5b`. Reusing `validate_k8s_name` (RFC 1123 label, 63 chars, no dots) would have rejected them and reintroduced the exact bug reported in #232, I verified this before changing approach:

```
# with validate_k8s_name (rejected approach)
delete_runtime("torchtune-llama3.2-1b", confirmed=True)
  -> False  VALIDATION_ERROR  "name must be lowercase alphanumeric with hyphens"
```

So this uses the subdomain form (RFC 1123 subdomain, 253 chars, dots allowed between labels).

### Relationship to #233

#233 introduces a `validate_runtime_name` with the same name, regex and length limit for the discovery tools. I matched its signature and semantics deliberately so the two converge on one shared helper instead of conflicting. Whichever merges first, the other becomes a small rebase, happy to rebase onto #233, or to pull the helper into its own PR first if you would rather land it once and have both build on it.

### Scope of the claim

The Kubernetes client URL-encodes path segments and the API server would reject most of these names itself, so this is defence-in-depth and contract consistency, **not** a demonstrated privilege escalation, and I am not claiming one. The concrete defects fixed are:

1. The three tools disagreed with every sibling trainer module on identical input.
2. `success: True` was returned for names that cannot reference a real object.
3. An empty name reached a cluster-scoped delete call.

These tools are `platform-admin` persona only, which limits exposure.

## Related Issue

Fixes #235

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

### Tests added: 33 total

`kubeflow_mcp/trainer/api/platform_test.py` (co-located, using the existing `TestCase` harness):

- 10 parametrised invalid names × all three tools → `VALIDATION_ERROR`, asserting the K8s API was **not** called (`delete_cluster_custom_object`, `patch_cluster_custom_object`, `create_cluster_custom_object` all unasserted).
- 4 parametrised valid names, including both real torchtune runtimes, → still accepted.

`kubeflow_mcp/core/security_test.py`:

- 5 valid subdomain forms accepted, 12 invalid forms rejected (empty, whitespace, uppercase, underscore, space, `../escape`, `a/b`, leading/trailing hyphen, leading/trailing dot, double dot).
- Boundary: 253 chars accepted, 254 rejected.

### Falsifiability check

Reverted the three guards and re-ran the platform suite to confirm the tests actually detect the defect rather than describing the new code:

```
without the fix:  10 failed, 15 passed
with the fix:     25 passed
```

### Before / after

Invalid names (mocked CustomObjects API, `confirmed=True`):

| name | before | after |
|---|---|---|
| `""` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"   "` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"UPPER-Case"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"bad_underscore"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"name with space"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"../escape"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"name/slash"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| `"-leading"` / `"trailing-"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |
| 300 × `"a"` | `success: True`, API called | `VALIDATION_ERROR`, API not called |

Valid names, no regression, verified across all three tools:

| name | delete | patch | create |
|---|---|---|---|
| `torchtune-llama3.2-1b` | ok | ok | ok |
| `torchtune-qwen2.5-1.5b` | ok | ok | ok |
| `torch-distributed` | ok | ok | ok |
| `r1` | ok | ok | ok |

The preview path (`confirmed=False`) validates too, so an invalid name is rejected before the agent is shown a preview it could then confirm.

### Suite and tooling

- `594 passed, 8 skipped` (561 on `main` before this branch).
- `uv lock --check`, `ruff check`, `ruff format --check` all clean (78 files formatted).
- Tool schema snapshot passes and the baseline file is unchanged: no tool signature changed, only a validation branch inside three existing tools.
- Coverage: `trainer/api/platform.py` 44% → 48%, `core/security.py` 92%.

### Existing tests reviewed for conflicts

`test_patch_runtime_invalid_runtime_name_returns_not_found` uses `"missing-runtime"`, a *valid* name that does not exist, so it still exercises the 404 path and passes unchanged. No existing test asserted that a malformed name succeeds.


kubeflow_mcp/core/security.py
```diff
@@ -52,6 +56,36 @@ def validate_k8s_name(name: str, field: str = "name") -> ToolError | None:
...
+        return ToolError(
+            error=f"{field} too long (max {MAX_SUBDOMAIN_LENGTH})",
+            error_code=ErrorCode.VALIDATION_ERROR,
+        )
+
+    if not K8S_SUBDOMAIN_PATTERN.match(name):
+        return ToolError(
+            error=f"{field} must be lowercase alphanumeric with hyphens or dots",
+            error_code=ErrorCode.VALIDATION_ERROR,
+            details={"value": name, "pattern": K8S_SUBDOMAIN_PATTERN.pattern},
```
abhijeet-dhumal: Please validate each dot-separated label is ≤63 characters. The current total-length check accepts oversized DNS labels, which pass preview validation but are rejected by Kubernetes. Add a regression test for a 64-character label.

abhishekKokadwar: Good catch, fixed in d0c79c3.

Each dot-separated label is now checked against `MAX_NAME_LENGTH` (63) after the pattern match, since RFC 1123 caps labels independently of the 253-character total. `64*"a"` previously passed here and was only rejected once it reached the API server.

Regression tests added in `security_test.py`: a 64-character label alone, leading, trailing and repeated, plus the 63-character boundary in both the single-label and three-label forms. I also corrected `test_accepts_name_at_253_chars`, which used 253 consecutive characters and would have failed under the new rule; it now builds a valid 253-character name from four labels.

Full suite: 599 passed, 8 skipped.

kubeflow_mcp/core/security.py
```diff
@@ -52,6 +56,46 @@ def validate_k8s_name(name: str, field: str = "name") -> ToolError | None:
...
+            error_code=ErrorCode.VALIDATION_ERROR,
+        )
+
+    if len(name) > MAX_SUBDOMAIN_LENGTH:
+        return ToolError(
+            error=f"{field} too long (max {MAX_SUBDOMAIN_LENGTH})",
+            error_code=ErrorCode.VALIDATION_ERROR,
+        )
+
+    if not K8S_SUBDOMAIN_PATTERN.match(name):
```
abhijeet-dhumal: Please use fullmatch() or \Z here. re.match() with $ accepts trailing newlines, allowing values such as runtime\n to pass validation. Add a regression test for this case.

abhishekKokadwar: Fixed in 72fee9c. Both patterns now use `fullmatch()`.

You are right that `$` matches before a trailing newline, so `"runtime\n"` passed validation and carried the newline through to the API server.

`validate_k8s_name` had the same flaw, so I applied `fullmatch()` there too rather than only to the runtime validator. That one is used by the discovery, monitoring, lifecycle and training tools, where `"job\n"` was accepted the same way. Happy to split that into a separate PR if you would rather keep this one scoped to the runtime tools.

Regression tests added for both validators: trailing `\n`, `\r\n`, and repeated newlines. Reverting `fullmatch()` back to `match()` fails exactly those 4 tests, so they pin the behaviour.

Full suite: 606 passed, 8 skipped.

abhijeet-dhumal:
Thanks @abhishekKokadwar !
/lgtm
/approve

PR #237, opened 2026-09-13
fix(core): withhold _meta.next when a response reports blockers

### LINKED ISSUE #234: fix(core): _meta.next tells the agent to proceed when pre_flight reported blockers
### Description

`pre_flight()` and `check_compatibility()` return `success: True` with the verdict inside `data` (`compatible: false`, `blockers: [...]`). `_inject_meta()` in `kubeflow_mcp/core/server.py` only skips hint injection when the *envelope* carries an error:

```python
if "error" in result or "error_code" in result:
    return result
```

A blocked compatibility report passes that guard, so the static entry from `TOOL_NEXT_HINTS` (`kubeflow_mcp/common/constants.py`) is attached anyway:

- `pre_flight` → `"Proceed to list_runtimes() to find available runtimes"`
- `check_compatibility` → `"Call get_cluster_resources() or use pre_flight() for full check"`

This contradicts the tool's own output and the server instructions, which say *"If blockers returned, STOP and inform user."* `pre_flight` already early-returns `next_steps: ["Blocker: ..."]` in `planning.py` specifically to halt the agent, and then `_meta.next` overrides that with an instruction to continue.

Two guidance channels in the same response disagree. `_meta` exists for clients that do not consume server instructions or MCP resources (Ollama, custom agents), which are exactly the clients with no second source to correct it.

### Steps to Reproduce

No cluster required. The *verdict*, not connectivity, is what triggers this, so this mocks a **reachable** cluster that is simply below the minimum supported Kubernetes version:

```python
import asyncio, json
from unittest.mock import MagicMock, patch
from fastmcp import Client
from kubeflow_mcp.core.server import create_server


def crd(served):
    versions = []
    for name in served:
        v = MagicMock()
        v.name = name
        v.served = True
        versions.append(v)
    obj = MagicMock()
    obj.spec.versions = versions
    return obj


async def run(tool, k8s_version, served):
    ver = MagicMock()
    ver.get_code.return_value = MagicMock(git_version=k8s_version)
    ext = MagicMock()
    ext.read_custom_resource_definition.return_value = crd(served)
    core = MagicMock()
    core.list_node.return_value = MagicMock(items=[])

    with (
        patch("kubeflow_mcp.common.utils.get_version_api", return_value=ver),
        patch("kubeflow_mcp.common.utils.get_custom_objects_api", return_value=MagicMock()),
        patch("kubeflow_mcp.common.utils.get_core_v1_api", return_value=core),
        patch("kubernetes.client.ApiextensionsV1Api", return_value=ext),
    ):
        server = create_server(clients=["trainer"], persona="platform-admin", mode="full")
        async with Client(server) as client:
            result = json.loads((await client.call_tool(tool, {})).content[0].text)

    data = result["data"]
    compat = data if "compatible" in data else data["compatibility"]
    print(tool, "| compatible:", compat["compatible"], "| blockers:", compat["blockers"])
    print("   _meta.next:", result.get("_meta", {}).get("next"))


asyncio.run(run("check_compatibility", "v1.21.0", ["v1alpha1"]))
asyncio.run(run("pre_flight", "v1.21.0", ["v1alpha1"]))
```

This also reproduces with no kubeconfig at all (`blockers` then names the unreachable API and the missing Trainer CRD), and with a CRD that serves only `v1beta1`.

### Expected Behavior

When a response reports blockers, `_meta` should not tell the agent to advance to the next workflow phase. `phase` remains accurate and can stay; `next` is the part that is wrong.

### Actual Behavior

Reproduced on `main` at 0f90ca6:

```
check_compatibility | compatible: False | blockers: ['Kubernetes v1.21.0 is below minimum v1.27']
   _meta.next: Call get_cluster_resources() or use pre_flight() for full check
pre_flight | compatible: False | blockers: ['Kubernetes v1.21.0 is below minimum v1.27']
   _meta.next: Proceed to list_runtimes() to find available runtimes
```

**Scope.** Affected, because the hint asserts a readiness the same payload denies:

- `pre_flight`
- `check_compatibility`

Deliberately **not** affected, verified:

- `health_check`: returns `status: "degraded"`, and its hint is already phrased conditionally (`"If degraded, check get_server_logs() for errors"`), so it stays correct.
- Tools that fail through the error envelope (`SDK_ERROR`, `KUBERNETES_ERROR`, and so on): the existing guard already suppresses their hints.
- A healthy environment: with Kubernetes 1.31 and `v1alpha1` served, `compatible: true` still receives its hint, so suppressing on blockers is not a regression in the normal path.

**Proposed fix.** One guard in `_inject_meta()`, the single point every tool response passes through, rather than per-tool handling in `planning.py`, which would leave future client modules (optimizer, hub, spark) to repeat the same mistake:

```python
def _is_blocked(result: dict[str, Any]) -> bool:
    """Report whether a successful response still carries blocking findings."""
    data = result.get("data")
    if not isinstance(data, dict):
        return False
    scopes = [data]
    nested = data.get("compatibility")   # pre_flight nests the compatibility report
    if isinstance(nested, dict):
        scopes.append(nested)
    for scope in scopes:
        if scope.get("compatible") is False or scope.get("blockers"):
            return True
    return False
```

…consulted after the existing error check, keeping `phase` and dropping only `next`.

Verified locally against this change: blocked cases lose `next`, the healthy control keeps it, `561 passed, 8 skipped` with no regressions, `make verify` clean, and the tool schema snapshot unchanged (`_meta` is response data, not part of the advertised schema).

**Questions before I open a PR.** `_meta` is a response contract, and `AGENTS.md` lists those under *"What NOT to Change Without Discussion"*, hence this issue first:

1. Is suppressing `next` on a blocked response acceptable, or would you prefer `_meta` keep its shape with `next` carrying explicit stop guidance instead (for example `"Blocked, resolve blockers before continuing"`)? The second keeps the key present for clients that read it unconditionally. Happy to implement either.
2. Should the generic `data.blockers` check apply to every client module, or stay narrowed to the two planning tools for now? The generic form is what lets optimizer/hub/spark inherit the behaviour, which is why I leaned that way.
3. Is a `data`-shape convention (`compatible` / `blockers` as the agreed "do not advance" signal) worth recording in `docs/CONVENTIONS.md` so new client modules opt in by default?

I would like to work on this if the approach looks right.

### MCP Server Version

0.1.1 (`main` at 0f90ca6)

### Python Version

3.11

### Kubernetes Version

n/a. Reproduces with a mocked API and with no kubeconfig

### MCP Client

Custom / Other

### References

- `AGENTS.md`: "What NOT to Change Without Discussion" (response contracts)
- Related but distinct: #217 / #218 concern hint *text* pointing at unregistered `trainer://` resource URIs; this issue is about whether a `next` hint should be emitted at all when the response reports blockers. Different mechanism, no overlap in files touched.

abhishekKokadwar:
## Description

`pre_flight()` and `check_compatibility()` return `success: True` with their verdict inside `data` (`compatible: false`, `blockers: [...]`). `_inject_meta()` only skipped hint injection when the *envelope* carried an error, so a blocked compatibility report still received its `TOOL_NEXT_HINTS` entry:

- `pre_flight` → `"Proceed to list_runtimes() to find available runtimes"`
- `check_compatibility` → `"Call get_cluster_resources() or use pre_flight() for full check"`

That contradicts the blockers the tool had just reported, the `next_steps: ["Blocker: ..."]` early return in `planning.py` that exists to halt the agent, and the server instruction *"If blockers returned, STOP and inform user."* `_meta` is consumed by clients that do not read server instructions or MCP resources (Ollama, custom agents), exactly the clients with no second source to correct it.

This reads the payload verdict in addition to the envelope. The workflow `phase` stays accurate and is preserved; only the advance hint is withheld.

The check lives in `_inject_meta`, the single point every tool response passes through, rather than in `planning.py`, so the optimizer, hub and spark modules inherit the behaviour instead of repeating the omission.

`health_check` is deliberately unaffected: it reports degradation without blockers, and its hint is already phrased conditionally (`"If degraded, check get_server_logs() for errors"`).

## Related Issue

Fixes #234

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

Eleven tests added in `kubeflow_mcp/core/server_test.py`. The two that assert the fix fail on `main` and pass here; the rest pin behaviour that must not change (hint present when compatible, empty `blockers` list treated as clean, unrelated payloads unaffected, `pre_flight`'s sibling `cluster`/`estimate`/`runtimes` sub-payloads not consulted, degraded `health_check` keeps its conditional hint, error envelope still short-circuits, non-dict passthrough).

To confirm the guard cannot over-reach, every trainer tool was called with a healthy mocked cluster and its real payload inspected for the two scopes the guard reads (`data` and `data.compatibility`):

| tool | `data.compatible` / `data.blockers` | `_is_blocked` |
|---|---|---|
| `check_compatibility` | `True` / `[]` | False |
| `pre_flight` | nested: `True` / `[]` | False |
| `get_cluster_resources`, `estimate_resources`, `list_runtimes`, `get_runtime`, `list_training_jobs`, `get_training_job`, `get_training_logs`, `get_training_events`, `health_check`, `get_server_logs` | absent | False |

Only the two planning tools populate those keys, so no other tool can lose its hint.

Manual verification through a real `fastmcp.Client` against a mocked *reachable* cluster below the minimum Kubernetes version. The bug is about the verdict, not connectivity:

Before:

```
check_compatibility | compatible: False | blockers: ['Kubernetes v1.21.0 is below minimum v1.27']
   _meta.next: Call get_cluster_resources() or use pre_flight() for full check
pre_flight | compatible: False | blockers: ['Kubernetes v1.21.0 is below minimum v1.27']
   _meta.next: Proceed to list_runtimes() to find available runtimes
```

After:

```
check_compatibility | compatible: False | blockers: ['Kubernetes v1.21.0 is below minimum v1.27']
   _meta.next: None
pre_flight | compatible: False | blockers: ['Kubernetes v1.21.0 is below minimum v1.27']
   _meta.next: None
```

A healthy control (Kubernetes 1.31, CRD serving `v1alpha1`) still receives its hint, so the normal path is unchanged.

Full suite: `572 passed, 8 skipped`. Coverage 74% → 75% overall, `core/server.py` 83% → 87%. `make verify` clean. The tool schema snapshot is unchanged, `_meta` is response data, not part of the advertised schema.

## Note on the response contract

`AGENTS.md` lists response contracts under *"What NOT to Change Without Discussion"*, so #234 raises the design question first and this PR implements the option I proposed there (withhold `next`). If you would rather keep the key present with explicit stop guidance, for example `"Blocked, resolve blockers before continuing"`, that is a small change to this diff and I am happy to switch.


abhijeet-dhumal:
Thanks @abhishekKokadwar, great catch!
/lgtm
/approve

PR #238, opened 2026-09-13
fix(core): set OTLP exporter timeout in seconds

### LINKED ISSUE #228: fix(core): OTLP exporter timeout is 2000 seconds instead of 2
### Description

`setup_tracing()` builds the exporter with `OTLPSpanExporter(..., timeout=2000)` right next
to `BatchSpanProcessor(exporter, export_timeout_millis=2000)`. The processor argument is
milliseconds, but the exporter's `timeout` is in seconds, so the exporter gets 2000 seconds
instead of 2.

With an unreachable collector, the exporter keeps retrying against that deadline, which
stalls span export and slows down process exit.

### Steps to Reproduce

1. Configure tracing against a collector that isn't running, e.g. `http://127.0.0.1:9`
2. Emit a span, then call `trace.get_tracer_provider().shutdown()`
3. Time it

### Expected Behavior

Shutdown gives up after roughly the intended 2 seconds.

### Actual Behavior

Shutdown took 30.0s in my run, retrying with backoff. Patching the exporter to
`timeout=2` brings it down to 1.2s. In opentelemetry-exporter-otlp-proto-http 1.43.0 the
exporter computes `deadline_sec = time() + self._timeout` and uses it as the requests
timeout, so the value is seconds.

### MCP Server Version

0.1.1

### Python Version

None

### Kubernetes Version

_No response_

### MCP Client

None

### References

_No response_

yats0x7:
## Description

`OTLPSpanExporter` reads its `timeout` in seconds (`float | None`, default 10). It was set to `2000`, on the line above `BatchSpanProcessor(exporter, export_timeout_millis=2000)`, which is the one that takes milliseconds. So the exporter was configured for 2000 seconds rather than the 2 the neighbouring line intends.

This shows up when the collector is unreachable. In opentelemetry-exporter-otlp-proto-http 1.43.0 the exporter computes `deadline_sec = time() + self._timeout`, passes the remainder as the requests timeout, and retries with backoff until that deadline runs out.

I measured `provider.shutdown()`, which runs at process exit through `atexit`, against a collector that was not listening:

| exporter timeout | shutdown returned after |
| --- | --- |
| 2000 (before) | 30.0s |
| 2 (after) | 1.2s |

## Related Issue

Fixes #228

## Checklist

## Testing

- [x] Unit tests

`test_setup_tracing_sets_exporter_timeout_in_seconds` asserts the exporter gets 2 and the batch processor gets 2000. It fails on main with `assert 2000 == 2`. Full suite locally: 562 passed, 8 skipped.


abhijeet-dhumal:
Thanks @yats0x7, good catch!
/lgtm
/approve

Issue #67, opened 2026-07-12
test(core): expand unit coverage for core MCP modules


abhijeet-dhumal:
Part of #9. To build on the test scaffold in #6.

Raise core module coverage and clear TODO(test) markers under tests/unit/core/ 

Scope

- dynamic_tools: registry, progressive/semantic search, keyword matching
- server: create_server, persona filtering, read-only mode, tool registration
- health and logging: health check, formatters, correlation IDs
- resources: URI resolution and MCP resource registration
- auth: JWT validation paths not yet covered
- tool_integrity: pin description checksum baseline
- config: malformed YAML, auth/OTEL env overrides

Out of scope

- Trainer tools (separate issue)
- Integration/e2e
- Benchmarks (#10 / #26)

Done when
- TODO(test) markers in tests/unit/core/ cleared or tracked
- make test-python green
- Meaningful coverage increase for kubeflow_mcp/core/

Prachi-Gupta2808:
/assign 

Will be working on this and ask if have any clarifying questions. Thank you

Prachi-Gupta2808:
Hi @abhijeet-dhumal should I wait for PR #6 to merge into main first, or would you be okay if I branch off feat/trainer-mcp-tests directly and rebase once #6 is merged?

abhijeet-dhumal:
> Hi [@abhijeet-dhumal](https://github.com/abhijeet-dhumal) should I wait for PR [#6](https://github.com/kubeflow/mcp-server/pull/6) to merge into main first, or would you be okay if I branch off feat/trainer-mcp-tests directly and rebase once [#6](https://github.com/kubeflow/mcp-server/pull/6) is merged?

That would be great, Thanks @Prachi-Gupta2808 for your interest in working on this! 🙌

PR #241, opened 2026-09-14
fix(core): stop bad execute_tool arguments from tripping the breaker

### LINKED ISSUE #239: fix(core): bad arguments to execute_tool trip the circuit breaker in dynamic mode
### Description

In progressive and semantic tool modes, `execute_tool` runs the target tool with `func(**args)` (`kubeflow_mcp/core/dynamic_tools.py`). When an agent passes a misspelled or unknown argument name, Python raises `TypeError`, which lands in the `except Exception` branch and calls `breaker.record_failure()`.

Five of those open the circuit breaker, since `failure_threshold` defaults to 5. From then on, correct calls to that tool are refused for the recovery timeout with a message saying the Kubernetes API may be degraded, even though the cluster is healthy and the only problem was a typo in the arguments.

The code already draws this distinction elsewhere. `is_infrastructure_error()` in `kubeflow_mcp/common/constants.py` states that validation errors are not infrastructure problems and should not trip the breaker, and the success path respects that. The exception path never consults it.

Full mode is unaffected, because FastMCP validates arguments against the tool schema before the function runs. Dynamic mode takes a raw `arguments` dict and skips that step, and agents do get argument names slightly wrong.

Binding the arguments with `inspect.signature(func).bind(**args)` before calling would let this return a validation error and leave the breaker alone.

### Steps to Reproduce

```python
from kubeflow_mcp.core import dynamic_tools as dt

def probe_tool(name: str) -> dict:
    """A fake read-only tool."""
    return {"success": True, "data": {"name": name}}

dt.init_dynamic_tools([probe_tool], {})
for _ in range(5):
    dt.execute_tool("probe_tool", {"nmae": "typo"})   # misspelled argument name

print(dt.execute_tool("probe_tool", {"name": "ok"}))  # correct call
```

In a real deployment this is an agent calling `execute_tool` five times against one tool with an argument name it got slightly wrong.

### Expected Behavior

A bad argument name is a client mistake, so it should return `VALIDATION_ERROR` and leave the breaker untouched. The correct call at the end should succeed.

### Actual Behavior

Each bad call returns `SDK_ERROR`, the breaker opens on the fifth, and the correct call is then refused:

```
bad-arg call 1: {'error': "probe_tool() got an unexpected keyword argument 'nmae'", 'error_code': 'SDK_ERROR', 'tool': 'probe_tool'}
bad-arg call 2: {'error': "probe_tool() got an unexpected keyword argument 'nmae'", 'error_code': 'SDK_ERROR', 'tool': 'probe_tool'}
bad-arg call 3: {'error': "probe_tool() got an unexpected keyword argument 'nmae'", 'error_code': 'SDK_ERROR', 'tool': 'probe_tool'}
bad-arg call 4: {'error': "probe_tool() got an unexpected keyword argument 'nmae'", 'error_code': 'SDK_ERROR', 'tool': 'probe_tool'}
bad-arg call 5: {'error': "probe_tool() got an unexpected keyword argument 'nmae'", 'error_code': 'SDK_ERROR', 'tool': 'probe_tool'}

Circuit breaker: CLOSED -> OPEN (failures=5)

valid call now: {'error': "Circuit breaker open for 'probe_tool' — K8s API may be degraded. Retries automatically after recovery timeout.", 'error_code': 'CIRCUIT_OPEN'}
```

### MCP Server Version

0.1.1 (main)

### Python Version

3.12

### References

#221 covers the breaker getting stuck in half-open. That is a separate problem in the same component.

yats0x7:
## Description

In progressive and semantic modes, `execute_tool` ran the tool with `func(**args)` and counted any exception as a breaker failure. A misspelled or missing argument raises `TypeError`, so five typos against one tool opened its breaker, and the next correct call came back `CIRCUIT_OPEN` saying the K8s API may be degraded. Full mode never hits this, because FastMCP checks arguments against the schema before the tool runs.

This binds the arguments to the tool's signature first and returns `VALIDATION_ERROR` when they don't fit. That lines up with `is_infrastructure_error()`, which already treats validation errors as a caller problem that shouldn't count against the breaker.

The check sits before `breaker.can_execute()` on purpose. In half-open state, `can_execute()` reserves a probe slot, and returning early after that would leave the slot used with nothing recorded. It's close to what #221 describes.

Exceptions raised inside a tool still count as failures, so real SDK and API errors trip the breaker the same way they did before.

## Related Issue

Fixes #239

## Checklist

## Testing

- [x] Unit tests

There was no test file for `dynamic_tools.py`, so this adds `kubeflow_mcp/core/dynamic_tools_test.py`. It only covers this bug and doesn't try to fill in the rest of the module, which #67 is tracking.

- unknown, extra and missing arguments return `VALIDATION_ERROR` and never call the tool
- repeated bad calls leave the breaker closed with `failure_count == 0`, and a correct call afterwards works
- a bad call while half-open doesn't use a probe slot
- a tool that raises still opens the breaker once it hits the threshold

On main, 5 of the 6 fail. The last one checks behaviour that shouldn't change, so it passes either way. Full suite locally: 580 passed, 8 skipped.


abhijeet-dhumal:
Thanks @yats0x7, great catch!
/lgtm
/approve

PR #250, opened 2026-09-16
fix: validate runtime names before Kubernetes operations

### LINKED ISSUE #249: fix: validate runtime names before Kubernetes operations
## Problem

The platform runtime tools in `kubeflow_mcp/trainer/api/platform.py` accept a Kubernetes runtime name but do not consistently validate it before processing the request.

Affected functions:

- `create_runtime()`
- `patch_runtime()`
- `delete_runtime()`

Invalid names such as `my runtime`, `INVALID_NAME`, or `runtime/name` can reach preview handling or Kubernetes API calls and produce unclear errors instead of the standard `VALIDATION_ERROR` response.

Other Trainer tools already use the existing `validate_k8s_name()` helper from `kubeflow_mcp/core/security.py`.

## Proposed Fix

Update `kubeflow_mcp/trainer/api/platform.py` to call `validate_k8s_name()` at the beginning of:

- `create_runtime()`
- `patch_runtime()`
- `delete_runtime()`

Validation should happen before patch/spec validation, preview handling, Kubernetes client initialization, and any Kubernetes API call.

## Tests

Add focused tests in `kubeflow_mcp/trainer/api/platform_test.py`.

Verify that:

- Invalid names return `VALIDATION_ERROR`.
- Kubernetes APIs are not called when validation fails.
- Valid names preserve existing behavior.
- Preview and confirmed requests validate names consistently.

## Acceptance Criteria

- All three runtime tools validate names at the tool boundary.
- Invalid Kubernetes names return `VALIDATION_ERROR`.
- Invalid names never reach the Kubernetes SDK.
- Existing valid-name behavior remains unchanged.
- Focused tests, linting, formatting, and the existing test suite pass.

pratik-naik003:
## Description

Adds Kubernetes runtime-name validation to the platform runtime tools.

The following functions now validate runtime names before preview handling or Kubernetes API calls:

- `create_runtime()`
- `patch_runtime()`
- `delete_runtime()`

Invalid names now return the standard `VALIDATION_ERROR` response.

## Testing

- [x] Added focused tests for all three runtime tools
- [x] Verified Kubernetes APIs are not called for invalid names
- [x] Platform tests pass (`14 passed`)
- [x] Ruff lint and format checks pass
- [x] Commit signed off

Fixes #249


abhishekKokadwar:
Heads up, and apologies for the collision: this overlaps #236, which merged as `ba5b9f7` about four hours after you opened this. `main` now has `validate_runtime_name(name)` at the top of all three of `create_runtime`, `patch_runtime` and `delete_runtime`, so the change here is already in place. Timing only, nothing you did wrong.

One detail worth carrying forward regardless of what happens to this PR, since it is the reason #236 did not use `validate_k8s_name`: that helper enforces the RFC 1123 *label* form, which excludes dots, and the torchtune runtimes Trainer ships have dots in their names.

```
name                        validate_k8s_name (this PR)   validate_runtime_name (main)
torchtune-llama3.2-1b       REJECT                        accept
torchtune-qwen2.5-1.5b      REJECT                        accept
torch-distributed           accept                        accept
```

`validate_k8s_name error: runtime name must be lowercase alphanumeric with hyphens`

So this version would make the two runtimes `fine_tune()` actually recommends undeletable and unpatchable, which is the same class of problem #232 reports for `get_runtime`. Runtime names are Kubernetes object names, so they take the subdomain form (dots allowed between labels, 253 total, each label capped at 63).

If it is useful, the review on #236 surfaced two more edge cases that are now covered in `validate_runtime_name` and are easy to miss:

- the 253 total length does not imply each dot-separated label is under 63, so `"a"*64` passed a total-length-only check and was then rejected by the API server
- `re.match` with a `$`-anchored pattern accepts a trailing newline, so `"runtime\n"` validated; it needs `fullmatch()`

Both apply to `validate_k8s_name` too and were fixed there in the same PR.

Since #249 is presumably closed by #236 now, this may just be closeable. If you would rather keep it open for something not covered, the gap I know of is `inspect_crd(name=...)` in the same file, which still takes a name without validating it.

PR #258, opened 2026-09-17
fix: use VALIDATION_ERROR for invalid model ID in estimate_resources

### LINKED ISSUE #257: fix: use VALIDATION_ERROR for invalid model ID in estimate_resources
Image

## Problem

`estimate_resources()` in `kubeflow_mcp/trainer/api/planning.py` (line 540-542) returns `ErrorCode.SDK_ERROR` for invalid model ID formats (e.g., `"random gibberish"`). It should return `VALIDATION_ERROR` instead.

AI agents interpret `SDK_ERROR` as a transient backend failure and retry the same bad input, instead of fixing it.

Other validations in the same function (`num_workers`, `batch_size`, `quantization`) correctly use `VALIDATION_ERROR`.

## Proposed Fix

Modify the error handler at line 531-544 to return `VALIDATION_ERROR` for format errors and `SDK_ERROR` for actual API/network failures. Preserve the existing `"did you mean"` suggestion logic.

## Acceptance Criteria

- Invalid model ID format returns `VALIDATION_ERROR`, not `SDK_ERROR`.
- API/network errors continue returning `SDK_ERROR`.
- Focused tests, linting, and full test suite pass.

pratik-naik003:
## Problem

`estimate_resources()` in `kubeflow_mcp/trainer/api/planning.py` (line 540-542) returns `ErrorCode.SDK_ERROR` for invalid model ID formats (e.g., `"random gibberish"`). It should return `VALIDATION_ERROR` instead.

AI agents interpret `SDK_ERROR` as a transient backend failure and retry the same bad input, instead of fixing it.

Other validations in the same function (`num_workers`, `batch_size`, `quantization`) correctly use `VALIDATION_ERROR`.

## Fix

Differentiate between format validation errors and actual API/network failures:
- **Format errors** (invalid model ID) → `VALIDATION_ERROR`
- **API/network failures** (HF Hub unreachable) → `SDK_ERROR` (unchanged)

The existing `"did you mean"` suggestion logic is preserved.

## Tests

Added two regression tests:
- `test_estimate_resources_returns_validation_error_for_invalid_format`
- `test_estimate_resources_returns_sdk_error_for_api_failure`

All 22 tests pass. Lint and format checks pass.

Closes #257


abhishekKokadwar:
I ran into this same `SDK_ERROR` classification from a different direction while looking at #259, so I pulled the branch and traced the failure shapes `_get_model_info_from_hf` can return. Sharing in case it saves review time. The direction here looks right to me, and the two added tests both pass.

One case that I think slips through: `_get_model_info_from_hf` has three distinct outcomes, and `startswith("Invalid")` only catches the first.

```
CASE 1: malformed id (fails _HF_MODEL_ID_RE)
   error: "Invalid HuggingFace model ID format: 'random gibberish'"
   startswith('Invalid'): True    -> VALIDATION_ERROR  (correct)

CASE 2: well-formed id, repo does not exist (RepositoryNotFoundError)
   error: '404 Client Error. Repository Not Found for url: ...'
   startswith('Invalid'): False   -> SDK_ERROR         (still misclassified)

CASE 3: genuine network/API failure
   result: {'error': 'Connection timed out'}
   startswith('Invalid'): False   -> SDK_ERROR         (correct)
```

Case 2 is `estimate_resources("invalid/nonexistent-model-xyz")`, which is the plain typo case and probably the most common way a user gets here. The code already treats it as user input rather than a backend problem: the `except RepositoryNotFoundError` branch in `planning.py` is commented "a well-formed id but no such repo (usually a typo)" and it attaches the same "did you mean" suggestions as case 1. So the reasoning in your description, that agents retry `SDK_ERROR` instead of fixing the input, applies to case 2 at least as strongly.

Two options, both small:

1. Also treat the not-found branch as validation. It is the branch that already builds suggestions, so the intent is unambiguous there.
2. Rather than matching on the message text, have `_get_model_info_from_hf` tag which branch produced the error, for example `{"error": ..., "kind": "format" | "not_found" | "api"}`, and let `estimate_resources` map `kind` to the error code. That avoids the classification breaking if anyone rewords the message, and `"Invalid"` is a fairly easy prefix to collide with later.

Option 2 is more code but removes the string dependency. Either way it would be worth a third test for the not-found case, since that is the one currently uncovered.

Repro for case 2, no cluster or network needed:

```python
from unittest.mock import patch
from huggingface_hub.errors import RepositoryNotFoundError
from kubeflow_mcp.trainer.api.planning import _get_model_info_from_hf

class R:  # minimal response object for the exception
    status_code, headers, text, request = 404, {}, "Repository Not Found", None
    url = "https://huggingface.co/api/models/invalid/nonexistent-model-xyz"

with patch("huggingface_hub.model_info",
           side_effect=RepositoryNotFoundError("404 Client Error. Repository Not Found for url: ...",
                                               response=R())):
    print(_get_model_info_from_hf("invalid/nonexistent-model-xyz"))
```

Not a maintainer, so treat this as a suggestion rather than a blocker.

PR #262, opened 2026-09-18
fix(core): report the package version in serverInfo

### LINKED ISSUE #261: bug: serverInfo.version reports the FastMCP version instead of kubeflow-mcp's
### Description

`initialize` returns `serverInfo.version` as `3.4.4`, which is the installed FastMCP version. kubeflow-mcp itself is `0.1.1`, as in `kubeflow_mcp.__version__` and `server.json`. Clients and debug output show the wrong version. I'll open a PR for this.

### Steps to Reproduce

1. Run `kubeflow-mcp serve --transport stdio`
2. Send `initialize` from any MCP client
3. Read `serverInfo.version` in the result

### Expected Behavior

`0.1.1`, matching `kubeflow_mcp.__version__`.

### Actual Behavior

`3.4.4`.

### MCP Server Version

main (467348f)

### Python Version

3.12

### Kubernetes Version

n/a, no cluster needed

### MCP Client

a raw stdio client and FastMCP's in-memory `Client`

Ahmad-Faraj:
## Description

`create_server()` built `FastMCP` without a version, so `initialize` reported FastMCP's own (`3.4.4`) as `serverInfo.version`. This passes `kubeflow_mcp.__version__`, so it reports `0.1.1`.

## Related Issue

Fixes #261

## Checklist

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests

`test_server_info_reports_package_version` runs the handshake through FastMCP's in-memory `Client` and checks the reported version. It fails on main with `'3.4.4' == '0.1.1'`.


abhijeet-dhumal:
@copilot review

abhijeet-dhumal:
@Ahmad-Faraj pyproject.toml still allows fastmcp>=2.0.0, but the version constructor argument was added in FastMCP 2.9.0. 
Could we either raise the minimum dependency to fastmcp>=2.9.0 and update uv.lock, or use a compatibility fallback here? 
Otherwise installations resolving FastMCP 2.0–2.8 may fail at startup.
WDYT? @jaiakash @Krishna-kg732

Ahmad-Faraj:
@abhijeet-dhumal went with the floor. `create_server()` already calls `add_middleware`, which also needs 2.9.0, so 2.0 to 2.8 already fail at startup on main. Bumped to `fastmcp>=2.9.0` and updated `uv.lock`.

abhijeet-dhumal:
Thanks @Ahmad-Faraj, great catch!
/lgtm
/approve

